PR #5, opened 2021-01-30
Adding display_override to incubations spec

dmurph:
@mgiuca  / @marcoscaceres , I'm trying to figure out how to structure this document - this is a first draft of adding display_override to this spec, assuming the editorial changes of this pull request go through:
https://github.com/w3c/manifest/pull/949 

What do you think? Should I do this a different way?

Thanks!
Daniel

### DIFF
--- a/index.html
+++ b/index.html
@@ -53,9 +53,9 @@
       <p>
         The following shows a [=manifest=] that prefers the
         <code>minimal-ui</code> <a data-cite=
-        "appmanifest#dfn-display-mode">display modes</a> to
-        <code>standalone</code>, but if minimal-ui isn't supported, fall back
-        to standalone as opposed to browser.
+        "appmanifest#dfn-display-mode">display mode</a> over
+        <code>standalone</code>, but if <code>minimal-ui</code> isn't supported, falls back
+        to <code>standalone</code> as opposed to <code>browser</code>.
       </p>
       <pre class="example json" title="Advanced display usage manifest">
         {
--- a/index.html
+++ b/index.html
@@ -8,17 +8,22 @@
       // See https://github.com/w3c/respec/wiki/ for how to configure ReSpec
       var respecConfig = {
         specStatus: "CG-DRAFT",
-        shortName: "beforeinstallprompt",
         editors: [{
           name: "Marcos Cáceres"
-        }]
+        }],
+        github: "WICG/manifest-incubations",
+        shortName: "manifest-incubations",
+        xref: {
+          specs: ["appmanifest"],
+          profile: "web-platform",
+        },
       };
     </script>
   </head>
   <body>
     <section id="abstract">
       <p>
-        This specification does neat stuff.
+        Feature specifications for manifest extensions & incubations.
       </p>
     </section>
     <section id="sotd">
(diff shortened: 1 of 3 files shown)


index.html
```diff
@@ -8,17 +8,22 @@
       // See https://github.com/w3c/respec/wiki/ for how to configure ReSpec
       var respecConfig = {
         specStatus: "CG-DRAFT",
-        shortName: "beforeinstallprompt",
         editors: [{
           name: "Marcos Cáceres"
```
mgiuca, line 12: Change to yourself. (Since this was empty before, I think @marcoscaceres won't mind not being an editor; we can always add back later.)

index.html
```diff
@@ -8,17 +8,22 @@
...
+          profile: "web-platform",
+        },
       };
     </script>
   </head>
   <body>
     <section id="abstract">
       <p>
-        This specification does neat stuff.
+        Feature specifications for manifest extensions & incubations.
```
mgiuca, line 26: Perhaps a little more context here to explain why things might go in here and not the main spec.

index.html
```diff
@@ -34,5 +39,98 @@ <h2>Introduction</h2>
...
+      <h2>
+        <dfn data-dfn-for="">WebAppManifestIncubations</dfn> dictionary
+      </h2>
+      <pre class="idl">
+          partial dictionary WebAppManifest {
+            sequence&lt;DisplayModeType&gt; display_override;
+          };
+      </pre>
+      <p>
+        This is an expansion of {{WebAppManifest}}.
```
mgiuca, line 111: Maybe s/expansion/extension (I never know which of those words to use -- remember Extended Memory vs Expanded Memory? I can still never remember which is which.) but "extension" is consistent with "extension point" terminology.

index.html
```diff
@@ -34,5 +39,98 @@ <h2>Introduction</h2>
...
+        their prefered <a data-cite="appmanifest#dfn-display-mode">display mode
+        </a>.
+      </p>
+      <p>
+        The <dfn>extension point for evaluating display_override</dfn> is given
+        by the following algorithm addition to the
+        <a data-cite="appmanifest#dfn-steps-for-determing-the-web-app-s-chosen-display-mode">
+        steps for determining the web app's chosen display mode</a>. The
+        following algorithm is executed at the
+        <a data-cite="appmanifest#dfn-extension-point">extension point</a>:
```
mgiuca, line 84: I find this paragraph a little hard to read.

I think this should use the text in [Web Share Target](https://w3c.github.io/web-share-target/level-2/#extension-to-the-web-app-manifest):

"The following steps are added to the `<a>`extension point`</a>` in `<a>`determining the web app's chosen display mode`</a>`:"

mgiuca (approved):
Hi Daniel,

Yep looks good! Approved. Did you want me to review w3c/manifest#949 again as well?

dmurph:
Thanks! Yes can you take a look at w3c/manifest#949? It's a new pull request to preserve the old one

index.html
```diff
@@ -1,38 +1,140 @@
...
+    <section class="informative">
+      <h2>
+        Incubation Usage Examples
+      </h2>
+      <p>
+        The following shows a [=manifest=] that prefers the
+        <code>minimal-ui</code> <a data-cite=
+        "appmanifest#dfn-display-mode">display modes</a> to
+        <code>standalone</code>, but if minimal-ui isn't supported, fall back
+        to standalone as opposed to browser.
```
tomayac, line 58:
```suggestion
        "appmanifest#dfn-display-mode">display mode</a> over
        <code>standalone</code>, but if <code>minimal-ui</code> isn't supported, falls back
        to <code>standalone</code> as opposed to <code>browser</code>.
```

PR #8, opened 2021-03-17
Adding share_target

dmurph:
***
<a href="https://pr-preview.s3.amazonaws.com/dmurph/manifest-incubations/pull/8.html" title="Last updated on Mar 17, 2021, 8:29 PM UTC (03bd79b)">Preview</a> | <a href="https://pr-preview.s3.amazonaws.com/WICG/manifest-incubations/8/e685d3c...dmurph:03bd79b.html" title="Last updated on Mar 17, 2021, 8:29 PM UTC (03bd79b)">Diff</a>

### DIFF
--- a/.pr-preview.json
+++ b/.pr-preview.json
@@ -1,0 +1,4 @@
+{
+    "src_file": "index.html",
+    "type": "respec"
+}
--- a/README.md
+++ b/README.md
@@ -1,7 +1,9 @@
 
 # Specification 'manifest-incubations'
 
-There's nothing here yet, but here's a brief description of what this repo is for.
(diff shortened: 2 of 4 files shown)


kenchris:
This seems pretty established, why don't we just add this to the main spec

dmurph:
It's there in the 'incubations' sections - my plan is to move those incubations out into this spec:
https://www.w3.org/TR/appmanifest/#incubations

Do more browsers support share_target now? IF so I can try to add it back to the main spec.

kenchris:
Support is a bit like the rest of web app manifest. Firefox only does mobile and Safari only supports the basics.

Edge and Chrome supports share targets, and I believe it is not entirely the same implementation due to OS integration

PR #9, opened 2021-03-17
Beforeinstallprompt

dmurph:
@mgiuca can you PTAL at this (I have to clean up the forked commit chain here, it has gotten really messy), but I'm interested in any ideas you have about putting the spec text into here.

I wonder if we need all of this, or if we can trim it down? But maybe it's nice to have everything.

I think the only stuff left in manifest spec is installability & installability signals, so I did one change on the Installation process to link to that, and removed the appinstalled event.

This doesn't link properly at all - but interested in your thoughts about how this might look (before I start manually data-citing everything).

***
<a href="https://pr-preview.s3.amazonaws.com/dmurph/manifest-incubations/pull/9.html" title="Last updated on Mar 19, 2021, 9:31 PM UTC (9983aa1)">Preview</a> | <a href="https://pr-preview.s3.amazonaws.com/WICG/manifest-incubations/9/4714290...dmurph:9983aa1.html" title="Last updated on Mar 19, 2021, 9:31 PM UTC (9983aa1)">Diff</a>

### DIFF
--- a/index.html
+++ b/index.html
@@ -53,9 +53,9 @@
       <p>
         The following shows a [=manifest=] that prefers the
         <code>minimal-ui</code> <a data-cite=
-        "appmanifest#dfn-display-mode">display modes</a> to
-        <code>standalone</code>, but if minimal-ui isn't supported, fall back
-        to standalone as opposed to browser.
+        "appmanifest#dfn-display-mode">display mode</a> over
+        <code>standalone</code>, but if <code>minimal-ui</code> isn't supported, falls back
+        to <code>standalone</code> as opposed to <code>browser</code>.
       </p>
       <pre class="example json" title="Advanced display usage manifest">
         {
--- a/index.html
+++ b/index.html
@@ -3,7 +3,7 @@
   <head>
     <meta charset="utf-8">
     <title>
-      Spec proposal
+      Manifest Incubations
     </title>
     <script src="https://www.w3.org/Tools/respec/respec-w3c" class=
     "remove"></script>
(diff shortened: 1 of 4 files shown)


mgiuca:
Hi Daniel!

Sorry it's taken so long for me to look at this.

I think this basically looks fine as is (modulo linking things correctly back to the Manifest spec). (But I am biased as I wrote most of it.)

> removed the appinstalled event

Is there a reason you removed it? From memory, we had a line in "steps to install the web application" that said something to the effect of "fire an Event named `"appinstalled"` at _window_" (that's all we need to say; no need to define an `AppInstalledEvent` type since it has no custom members or methods). Can you put that line back? (Or am I misremembering?)

You still mention the `window.onappinstalled` event handler, and have an example of handling the event both ways, which currently doesn't make sense since there's no code that fires an event called `"appinstalled"`.

I'd like it back, since we support both events in Chrome (& presumably Edge).

> I wonder if we need all of this, or if we can trim it down? But maybe it's nice to have everything.

It could possibly be trimmed down. I was thinking maybe we didn't need the distinction between the three types of install prompt (manual, automated and site-triggered), but looking at it, I think those are all treated semantically differently. (For example, manual doesn't trigger BIP at all, automated requires BIP to fire right before it, and site-triggered requires BIP to fire, then the site can opt into triggering it.) So I think we need all of this.

Another option is to delete automated prompts entirely -- since I don't think any current implementations do it. (i.e. remove this from being a normative possibility.) That's in there because it was originally the *only* behaviour that Chrome did. But we took that out, and the current spec text (before it was removed) gives UAs the option of either automated or site-triggered. But Chrome for Desktop and Android (or any other browser that implements this) _I believe_, do not use automated prompts. We could just remove that whole possibility, and add it back if an implementor needs it.

mgiuca (approved):
Comments above.

lgtm with appinstalled event firing back. Or, if you want to remove automated install prompts, feel free to ping for another review.

index.html
```diff
@@ -129,5 +129,383 @@ <h2 data-dft-for="">
...
+        <ul>
+          <li>An end-user can <dfn data-lt="manual installation">manually</dfn>
+          trigger the installation process through the user agent's
+            <abbr title="User Interface">UI</abbr>, directly invoking the steps
+            to <a>present an install prompt</a>.
+          </li>
+          <li>The installation process can occur through an <dfn>automated
+          install prompt</dfn>: that is, a UI that the user agent presents to
+          the user when, for instance, there are sufficient <a data-cite=
+          "appmanifest#installability-signals">installability signals</a> to
```
marcoscaceres, line 153: I removed the "installability signals" from the other spec, as they turned out to be somewhat Chrome specific. So it might be good to just say:

```suggestion
          the user when, for instance, there are sufficient signals to
```

index.html
```diff
@@ -129,5 +129,383 @@ <h2 data-dft-for="">
...
+          warrant <a data-cite="appmanifest#dfn-installed">installation</a> of
```
marcoscaceres, line 154: General note: you should never need to use `data-cite=` for any W3C spec. If a term is not available via xref, it needs to be `data-export=""`ed from the other spec.

index.html
```diff
@@ -129,5 +129,383 @@ <h2 data-dft-for="">
...
+        <p>
+          Installation events and supporting the {{BeforeInstallPrompt}} is
+          OPTIONAL.
+        </p>
+        <p>
+          DOM events <a>fired</a> by this specification use the
+          <dfn>application life-cycle task source</dfn>.
+        </p>
+        <section data-dfn-for="BeforeInstallPromptEvent" data-link-for=
+        "BeforeInstallPromptEvent">
```
marcoscaceres, line 253:
```suggestion
        <section data-dfn-for="BeforeInstallPromptEvent">
```

dmurph:
Apologies, I didn't mean to close this but I can't seem to reverse it. Pull request is here:
https://github.com/WICG/manifest-incubations/pull/12

PR #10, opened 2021-05-05
new_note_url parsed relative to manifest_url

phoglenix:
Clarify that new_note_url should be parsed with manifest_url as the base URL, like all other URLs in the manifest.

***
<a href="https://pr-preview.s3.amazonaws.com/phoglenix/manifest-incubations/pull/10.html" title="Last updated on Jun 28, 2021, 8:49 AM UTC (e2c1edc)">Preview</a> | <a href="https://pr-preview.s3.amazonaws.com/WICG/manifest-incubations/10/861f7ed...phoglenix:e2c1edc.html" title="Last updated on Jun 28, 2021, 8:49 AM UTC (e2c1edc)">Diff</a>

### DIFF
--- a/index.html
+++ b/index.html
@@ -134,31 +134,44 @@
         <code><dfn>note_taking</dfn></code> member
       </h2>
       <p>
-        The <code>note_taking</code> member of the <a data-cite=
+        The `note_taking` member of the <a data-cite=
         "appmanifest#web-application-manifest">Web Application Manifest</a> is
-        an <a data-cite="appmanifest#dfn-object">object</a> that contains
-        information related to note-taking. It has the following members:
+        an [=ordered map=] that contains information related to note-taking. It
+        has the following members:
       </p>
       <ul>
         <li>[=note_taking/new_note_url=]
         </li>
       </ul>
+      <p>
+        A user agent MAY use these members to treat the web application
+        differently as an application with note-taking capabilities (e.g.,
+        integrate with operating system note-taking surfaces).
+      </p>
       <section>
         <h3>
           <code><dfn data-dfn-for="note_taking">new_note_url</dfn></code> member
         </h3>
         <p>
-          The [=note_taking=] <code>new_note_url</code> member is a <a>URL</a>
-          to open when the user wants to take a new note using the web
-          application (e.g., from an operating system shortcut icon or keyboard
-          shortcut). Defining this member indicates that the web application is
-          a note-taking application.
+          The [=note_taking=] `new_note_url` member is a [=string=] that
+          represents the <a data-cite="url#concept-url">URL</a> the developer
+          would prefer the user agent load when the user wants to take a new
+          note using the web application (e.g., from an operating system
+          shortcut icon or keyboard shortcut).
         </p>
         <p>
-          The <code>new_note_url</code> member is parsed with
-          <a data-cite="appmanifest#dfn-manifest-url">manifest URL</a> as the
-          base URL and must be [=manifest/within scope=] of the manifest.
+          The `new_note_url` member is purely advisory,
+          and a user agent MAY <a data-cite="appmanifest#dfn-ignore">ignore</a>
+          it or provide the end-user the choice to modify it or to not use it.
         </p>
+        <aside class="note">
+          <p>
+            The `new_note_url` member is parsed with
+            <a data-cite="appmanifest#dfn-manifest-url">manifest URL</a> as the
+            base URL and is ignored if not [=manifest/within scope=] of the
+            manifest.
+          </p>
+        </aside>
       </section>
       <section class="informative">
         <h3>
@@ -182,6 +195,67 @@
             }
           }
         </pre>
+      </section>
+      <section>
+        <h3>
+          Processing the `note_taking` member
+        </h3>
+        <p>
+          To <dfn>process the `note_taking` member</dfn>, given [=ordered map=]
+          |json:ordered map|, [=ordered map=] |manifest:ordered map|, [=URL=]
+          |manifest_URL:URL|, run the following during the
+          <a data-cite=
+          "appmanifest#dfn-processing-extension-point-of-web-manifest">extension
+          point</a> in [=processing a manifest=]:
+        </p>
+        <ol class="algorithm">
+          <li>If |json|["note_taking"] does not [=map/exist=], return.</li>
+          <li>If the type of |json|["note_taking"] is not [=ordered map=],
+            return.
+          </li>
+          <li>Set |manifest|["note_taking"] to a new [=ordered map=].</li>
+          <li>If |json|["note_taking"]["new_note_url"] does not [=map/exist=],
+            return.
+          </li>
+          <li>If the type of |json|["note_taking"]["new_note_url"] is not
+            [=string=], return.
+          </li>
+          <li>Let |new_note_url:URL| be the result of [=URL Parser|parsing=]
+            |json|["note_taking"]["new_note_url"] with |manifest URL| as the
+            base URL.
+          </li>
+          <li>If |new_note_url| is failure, return.</li>
+          <li>If |new_note_url| is not [=manifest/within scope=] of |manifest
+            URL|, return.
+          </li>
+          <li>Set |manifest|["note_taking"]["new_note_url"] to |new_note_url|.
+          </li>
+        </ol>
+      </section>
+      <section>
+        <h3>
+          Launching the `new_note_url`
+        </h3>
+        <p>
+          To <dfn>launch the `new_note_url`</dfn>, given <a data-cite=
+          "appmanifest#dfn-processed-manifest">processed manifest</a>
+          |manifest:processed manifest|, run the following steps:
+        </p>
+        <ol>
+          <li>If |manifest|["note_taking"] does not [=map/exist=], return.</li>
+          <li>If |manifest|["note_taking"]["new_note_url"] does not
+            [=map/exist=], return.
+          </li>
+          <li>If the type of |manifest|["note_taking"]["new_note_url"] is not
+            [=URL=], return.
+          </li>
+          <li>Let |browsing context:Browsing Context| be the result of creating
+            a new [=top-level browsing context=].
+          </li>
+          <li>[=Navigate=] |browsing context| to resource
+            |manifest|["note_taking"]["new_note_url"].
+          </li>
+        </ol>
       </section>
     </section>
   </body>
--- a/index.html
+++ b/index.html
@@ -136,8 +136,8 @@
       <p>
         The `note_taking` member of the <a data-cite=
         "appmanifest#web-application-manifest">Web Application Manifest</a> is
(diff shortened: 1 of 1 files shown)


phoglenix:
@mgiuca PTAL

index.html
```diff
@@ -149,11 +149,15 @@ <h3>
...
         <p>
           The [=note_taking=] <code>new_note_url</code> member is a <a>URL</a>
-          [=manifest/within scope=] of the manifest, to open when the user
-          wants to take a new note using the web application (e.g., from an
-          operating system shortcut icon or keyboard shortcut). Defining this
-          member indicates that the web application is a note-taking
-          application.
+          to open when the user wants to take a new note using the web
+          application (e.g., from an operating system shortcut icon or keyboard
+          shortcut). Defining this member indicates that the web application is
```
mgiuca: I know this isn't the point of this PR, but I figure since we're making this text more formal, you could probably do it here at the same time.

This is pretty non-specific, especially since it doesn't define what a "note-taking application" is. I think the way to fix it is not to define "note-taking application", but to remove this sentence and replace it with an algorithm for launching the note URL. Similar to how Web Share Target defines [in an algorithm how to launch the share target URL](https://w3c.github.io/web-share-target/#launching-the-web-share-target), and file handling should do this too if it doesn't already.

Some roughly thought out potential spec text here:

"When the user agent needs to invoke the note-taking capabilities of an application with `WebAppManifest` _manifest_, run the following steps:

1. Let _browsing context_ be the result of creating a [new top-level browsing context](https://html.spec.whatwg.org/multipage/#top-level-browsing-context).
2. [Navigate](https://html.spec.whatwg.org/multipage/#navigate) _browsing context_ to a new [Request](https://fetch.spec.whatwg.org/#concept-request) whose method is `"GET"` and url is _manifest_`[new_note_url]`."

(I copied this from the last two steps of Web Share Target algorithm which should help get the Respec syntax right.

index.html
```diff
@@ -149,11 +149,15 @@ <h3>
...
+          a note-taking application.
+        </p>
+        <p>
+          The <code>new_note_url</code> member is parsed with
```
mgiuca: This isn't formal enough (e.g., doesn't actually say how to invoke the parser, doesn't say what to do iff it's not within scope).

"[process a shortcut](https://www.w3.org/TR/appmanifest/#dfn-process-a-shortcut)" is a good example of how to specify processing of a must-be-within-scope URL in the Manifest.

Quick note about the word "must": this is a special word (along with "required", "shall", "should", "recommended", "may" and "optional") defined in [RFC 2119](https://datatracker.ietf.org/doc/html/rfc2119) which has a specific meaning in normative language -- it imposes a requirement on the user agent which implements this spec. (So if we use it, it should always be capitalized, and should not appear in non-normative text.) The problem in this case is that you aren't imposing a requirement on the user agent, you're imposing a requirement on the manifest. We don't impose requirements on the documents that our specs process; rather, we say exactly what the user agent should do in the event that the document isn't valid. (So instead of saying "must be within scope", we give instructions for what to do if you find a URL that isn't within scope.)

So what should be done here:

1. I think put a separate section at the top of this document (above `display_override`) called "Extension to processing the manifest", which says "The following steps are added to the extension point in the steps for processing a manifest:" and then says "Process the `new_note` member, passing _json_ and _manifest_."
2. Here, create an algorithm called "To process the `new_note` member, given object _json_ and ordered map _manifest_:". In the body, invoke the URL parser on _json_`["new_note"]`; if it's failure, return (silently dropping the `new_note` member); otherwise if it isn't within scope, return (silently dropping); otherwise store the result in _manifest_`["new_note"]`. (Use "[process a shortcut](https://www.w3.org/TR/appmanifest/#dfn-process-a-shortcut)" as a guide.)

It turns out that (news to me) now every member needs an explicit processing step, not just ones with special requirements. Which means we actually need to add such a step for all of the other ones in peripheral specs, like `display_override`, `share_target`, etc. But let's leave that out for this PR.

mgiuca (changes requested):

index.html
phoglenix: Hmm, this paragraph is aiming to define the semantic meaning of the field as opposed to how to process it.
A web app author defining this field is indicating that the application is "a note-taking application", whatever that means, so the user agent _may_ want to treat the app differently (eg. show in list of note-taking apps). I'd be happy to say "an application with note-taking capabilities" instead of "a note taking application" if that is less of a stretch to use without a formal definition.
Perhaps I can model it on the [start_url definition](https://www.w3.org/TR/appmanifest/#start_url-member)?
I can put the algorithm for how to launch a new_note_url below.

index.html
phoglenix: I've moved the non-normative descriptive bits to a "Note" section and used MAY for the "treat the web application differently" and how to treat the new_note_url (similar to how start_url is defined).

I've added explicit processing steps including adding this as an extension to the manifest processing extension point. I've put it inline in the note_taking processing section instead of at the top of the doc so that each extension is independent of others.

index.html
```diff
@@ -134,27 +134,44 @@ <h2>
         <code><dfn>note_taking</dfn></code> member
       </h2>
       <p>
-        The <code>note_taking</code> member of the <a data-cite=
+        The `note_taking` member of the <a data-cite=
         "appmanifest#web-application-manifest">Web Application Manifest</a> is
-        an <a data-cite="appmanifest#dfn-object">object</a> that contains
-        information related to note-taking. It has the following members:
+        an [=ordered map=] that contains information related to note-taking. It
```
phoglenix: Not sure if it's correct to say the member is an infra ordered map (ie post processing) or a JSON object (ie. pre processing). WDYT?

mgiuca: I think I would say "object" (pre-processing), since you're describing the JSON data structure, not the internal representation after processing.

index.html
```diff
@@ -134,27 +134,44 @@ <h2>
...
+          The [=note_taking=] `new_note_url` member is a [=string=] that
+          represents the <a data-cite="url#concept-url">URL</a> the developer
+          would prefer the user agent load when the user wants to take a new
+          note using the web application (e.g., from an operating system
+          shortcut icon or keyboard shortcut).
+        </p>
+        <p>
+          The `new_note_url` member is purely advisory,
+          and a user agent MAY <a data-cite="appmanifest#dfn-ignore">ignore</a>
+          it or provide the end-user the choice to modify it or to not use it.
```
mgiuca: Sentence is hard to parse (mainly because of the nested "or"s).

(i.e. it could be read either as:

> (ignore it) or (provide the end-user the choice to modify it) or (to not use it)

or, as intended:

> (ignore it) or (provide the end-user the choice to (modify it) or (to not use it))

Perhaps reword as

"... MAY ignore it entirely, or provide the end-user the choice of whether to use it. The user agent MAY provide the end-user the choice to modify it."

index.html
```diff
@@ -179,6 +196,67 @@ <h3>
...
+          "appmanifest#dfn-processing-extension-point-of-web-manifest">extension
+          point</a> in [=processing a manifest=]:
+        </p>
+        <ol class="algorithm">
+          <li>If |json|["note_taking"] does not [=map/exist=], return.</li>
+          <li>If the type of |json|["note_taking"] is not [=ordered map=],
+            return.
+          </li>
+          <li>Set |manifest|["note_taking"] to a new [=ordered map=].</li>
+          <li>If |json|["note_taking"]["new_note_url"] does not [=map/exist=],
```
mgiuca: I think this should be put into a function ("processing the new note URL"). Because if we ever add a second member to `note_taking` that's independent of `new_note_url`, this will break since you have all these return statements on failure.

Instead, move this into a function, then all of those return statements will just return out of "processing the new note URL" rather than the whole processing function.

index.html
```diff
@@ -179,6 +196,67 @@ <h3>
...
+            return.
+          </li>
+          <li>If the type of |json|["note_taking"]["new_note_url"] is not
```
mgiuca: Aside: the next four lines appear so many times in the manifest spec, I think we should put it into a function "parse an in-scope URL", which parses a string as a URL relative to the manifest URL and returns failure if it isn't in scope.

Something for another time but just noting this down.

index.html
```diff
@@ -179,6 +196,67 @@ <h3>
...
+        <h3>
+          Launching the `new_note_url`
+        </h3>
+        <p>
+          To <dfn>launch the `new_note_url`</dfn>, given <a data-cite=
+          "appmanifest#dfn-processed-manifest">processed manifest</a>
+          |manifest:processed manifest|, run the following steps:
+        </p>
+        <ol>
+          <li>If |manifest|["note_taking"] does not [=map/exist=], return.</li>
```
mgiuca, line 263: Combine these two (or three?) into one statement with "or"s.

index.html
```diff
@@ -179,6 +196,67 @@ <h3>
...
+          <li>If |manifest|["note_taking"]["new_note_url"] does not
+            [=map/exist=], return.
+          </li>
+          <li>If the type of |manifest|["note_taking"]["new_note_url"] is not
+            [=URL=], return.
+          </li>
+          <li>Let |browsing context:Browsing Context| be the result of creating
+            a new [=top-level browsing context=].
+          </li>
+          <li>[=Navigate=] |browsing context| to resource
```
mgiuca, line 273: Another aside: I suspect all of these "create a new browsing context / navigate to URL" will need to change once we have the launch handler stuff in the spec.

So we should probably create a function called "launch a URL" which invokes that mechanism. Again, something for later but just writing it down.

mgiuca (commented):
> Hmm, this paragraph is aiming to define the semantic meaning of the field ...

The new paragraph is good.

> I've moved the non-normative descriptive bits to a "Note" section and ...

The note is good.

A few comments but I'm happy with this.

index.html
phoglenix: > "object" (pre-processing)

Per https://github.com/w3c/manifest/issues/984 I was assuming we will have an infra object, not a JSON one. But that hasn't landed so I will continue to pretend JSON objects are a thing that you can do operations on :stuck_out_tongue: 

Done.

> nested "or"s

Fixed.

index.html
phoglenix: Agreed.

index.html
phoglenix: Done.

index.html
phoglenix: Agreed.

index.html
phoglenix: I actually think it's a bit cleaner to keep them all separate. Adding "or"s to makes it a more complex statement and is more likely to flow over multiple lines without nice code formatting.
Keeping as-is if that's ok.

mgiuca: Ack.

index.html
mgiuca: https://github.com/w3c/manifest/issues/986

index.html
mgiuca: https://github.com/w3c/manifest/issues/986

mgiuca (approved):

PR #16, opened 2021-07-08
Add explainer for note_takng : new_note_url member

phoglenix:
Mostly using discussion at https://github.com/w3c/manifest/issues/965
View rendered markdown: https://github.com/phoglenix/manifest-incubations/blob/add-note-taking-explainer/note_taking-explainer.md

### DIFF
--- a/note_taking-explainer.md
+++ b/note_taking-explainer.md
@@ -17,6 +17,19 @@ new note. It wouldn't cover all the use-cases linked above (eg. voice assistant
 actually adding items to the note) but it covers the core flow of adding a new
 note.
 
+### Background
+
+Other web app APIs set a precedent of using a URL declared in the web app
+manifest to advertise capabilities and facilitate integrations with the OS:
+
+* [Web Share Target](https://w3c.github.io/web-share-target/)
+* [File Handling](https://github.com/WICG/file-handling/blob/main/explainer.md)
+* [Declarative Web Actions](https://github.com/slightlyoff/declarative_web_actions)
+
+[Protocol Handlers](
+https://html.spec.whatwg.org/multipage/system-state.html#dom-navigator-registerprotocolhandler)
+uses JavaScript for a similar purpose.
+
 ## Proposal
 
 This explainer proposes a new dictionary manifest entry `note_taking` with an
@@ -46,6 +59,36 @@ related functionality together. It also means note-taking can be specified
 largely orthogonally to other manifest specification changes, which helps to
 keep the manifest specification modular and manageable.
 
+### Future Considerations
+
+If the need arises in the future to add more note-taking functionality, this can
+easily be added to the `note_taking` member. For example (**not
+currently proposed**), if we want to be able to populate a note with contents
+from another app or add items to a list from voice input, we could add:
+
+```
+{
+  ... other manifest fields ...
+  "note_taking": {
+    "new_note_with_contents": {
+      "url": "/POST_url/add_note.html",
+      "params": {
+        "title": "name",
+        "text":  "description",
+        "url":   "link"
+      }
+    },
+    "add_item_to_note": {
+      "action": "/POST_url/add_item_to_note.html",
+      "params": {
+        "existing_note_id": "id",
+        "item_text":  "description",
+      }
+    }
+  }
+}
+```
+
 ## Alternatives Considered
 
 A common theme of alternatives is to make a generalised interface for web apps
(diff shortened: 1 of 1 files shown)


phoglenix:
@mgiuca PTAL

note_taking-explainer.md
```diff
@@ -0,0 +1,129 @@
...
+        "url":   "link"
+      }
+    }
+  }
+}
+```
+
+Where each separate integration has a particular schema of fields.
+
+However, these individual integration likely need to be specified individually
```
mgiuca: There's a bit of a missing link here: this document is roughly saying there are two alternatives:

- The basic `new_note_url` being proposed.
- The "over-engineered" integrations -> note-taking -> create mentioned here.

However, the perhaps obvious but instated fact is that the former can be extended with all of these features. For example, `note-taking` has `new-note-url` but that could also get an `add-item` section later, just like what you've got here. The fact that you _only_ mention `add-item` in the "alternatives considered" section implies that we won't be able to do it with your proposal.

It might be worth adding a "future considerations" section, saying that we could add `add-item` and maybe also a `new-note-with-contents` that would let a voice assistant take a new note and populate it with some spoken text.

note_taking-explainer.md
```diff
@@ -0,0 +1,129 @@
...
+extend and complicate the `Shortcuts` member. Additionally, Shortcuts are
+intended to be displayed to the user, while many tasks/capabilities should not
+be displayed in a shortcuts menu, eg. any parameterised or contextual tasks,
+non-task capabilities like [lock screen](https://github.com/WICG/lock-screen)
+support. It seems like extending `Shortcuts` this far would be significantly
+changing the semantics of the member.
+
+### A General Capability Member
+
+There are existing general capabilities approaches for other functionality of
```
mgiuca: I'm not exactly sure what the relevance of these four APIs is. They don't seem analogous to "integrations" & I would probably just take it out.

It might be worth mentioning them (or a subset of them) up above by way of precedent for the "manifest fields that have a URL that opens when you do something in the OS". (i.e., if anything, these four are a precedent for the proposed design, not the "integrations" alternative.)

note_taking-explainer.md
phoglenix: I tried to get that across when talking about the proposal but I agree it's not stated very obviously, so added a section as you suggested.

note_taking-explainer.md
phoglenix: To me, "Declarative Web Actions" feels like a more "general integrations" approach. But it also appears to be the one that's least implemented. Hmm, that might be telling us something.

I'll move them up above as background.

note_taking-explainer.md
```diff
@@ -17,6 +17,19 @@ new note. It wouldn't cover all the use-cases linked above (eg. voice assistant
...
+### Background
+
+Other web app APIs set a precedent of using a URL declared in the web app
+manifest to advertise capabilities and facilitate integrations with the OS:
+
+* [Web Share Target](https://w3c.github.io/web-share-target/)
+* [File Handling](https://github.com/WICG/file-handling/blob/main/explainer.md)
+* [Declarative Web Actions](https://github.com/slightlyoff/declarative_web_actions)
+
+[Protocol Handlers](
```
mgiuca: Actually there's a newer proposal https://web.dev/url-protocol-handler/ for doing protocol handlers in the manifest just like the above 3. It's a lot more relevant to this than the old HTML registerProtocolHandler, so delete this, and put the above link into the list.

note_taking-explainer.md
```diff
@@ -17,6 +17,19 @@ new note. It wouldn't cover all the use-cases linked above (eg. voice assistant
...
 note.
 
+### Background
+
+Other web app APIs set a precedent of using a URL declared in the web app
+manifest to advertise capabilities and facilitate integrations with the OS:
+
+* [Web Share Target](https://w3c.github.io/web-share-target/)
+* [File Handling](https://github.com/WICG/file-handling/blob/main/explainer.md)
+* [Declarative Web Actions](https://github.com/slightlyoff/declarative_web_actions)
```
mgiuca: I'm not sure it's worth mentioning this. I don't think Alex really pushed this proposal (I hadn't heard of it until I saw it here), so I don't think we can call it a "precedent".

note_taking-explainer.md
```diff
@@ -71,15 +114,8 @@ changing the semantics of the member.
...
-There are existing general capabilities approaches for other functionality of
-web apps/pages:
-
-* [Web Share Target](https://w3c.github.io/web-share-target/)
-* [File Handling](https://github.com/WICG/file-handling/blob/main/explainer.md)
-* [Protocol Handlers](https://html.spec.whatwg.org/multipage/system-state.html#dom-navigator-registerprotocolhandler)
-* [Declarative Web Actions](https://github.com/slightlyoff/declarative_web_actions)
-
-So perhaps we could do something similar with a more general-purpose field like:
+Perhaps we should use a more general-purpose field, where each separate
```
mgiuca: Rephrase this. "Perhaps we should" suggests you're considering it, but really this is something we've considered and discarded.

note_taking-explainer.md
phoglenix: Ok, removed.

note_taking-explainer.md
phoglenix: Ah cool. Done.

note_taking-explainer.md
phoglenix: Done.

mgiuca (approved):

PR #41, opened 2021-10-25
Add explainer for translations member

loubrett:


translations-explainer.md
```diff
@@ -0,0 +1,134 @@
+# Web App Translations
+
+Authors: Louise Brett (loubrett@google.com), Glen Robertson (glenrob@chromium.org)
+
+
```
phoglenix: nit: no need for double blank lines in markdown. I'm pretty sure any 1+ blank lines renders the same (fix here and below)

translations-explainer.md
```diff
@@ -0,0 +1,134 @@
...
+## Problem
+
+Currently the Web App Manifest has no direct support for localization so fields can only be provided in a single language. There are two recommended ways to translate the manifest suggested in [non-normative text in the spec](https://www.w3.org/TR/appmanifest/#internationalization):
+
+
+
+*   Using a different manifest URL for each language and having HTML point to the correct one.
+*   Dynamically serving a localized manifest from the same URL.
+
+There are several problem with these approaches:
```
phoglenix: nit "problems"

translations-explainer.md
```diff
@@ -0,0 +1,134 @@
...
+      "description": "Une application pour chiens",
+      "icons": [],
+      "screenshots": []
+    }
+  },
+}
+```
+
+
+When the user agent language matches a language present in the translations entry, those fields should be used instead of the top-level fields. If none of the provided languages match, the top-level fields should be used.
```
phoglenix: nit: try to avoid "should" because it implies spec language. Here, replace with "can"

translations-explainer.md
```diff
@@ -0,0 +1,134 @@
...
+
+Not every field in the Web App Manifest should be localizable. The fields which should be localizable and able to be provided in the translations are:
```
phoglenix: nit: as above, maybe "needs to be localizable" and "could be localizable"

phoglenix (changes requested):
Looking good, just some small things to fix

phoglenix (approved):
LGTM

PR #42, opened 2021-11-03
Add example for translating shortcuts

loubrett:


translations-explainer.md
```diff
@@ -54,6 +54,34 @@ Not every field in the Web App Manifest needs to be localizable. The fields whic
 *   `screenshots`
 *   `shortcuts`
 
+For complex fields such as `shortcuts` where only a subset of its fields can be translated (e.g. `name` but not `url`), only the fields to be translated should be redefined in `translations`. The ordering of the shortcuts will be used to match it’s translation. Therefore the number of shortcuts defined in the translations should match the number defined at the top level.
```
phoglenix: nit "match it’s translation" -> "match its translation"

phoglenix: "Therefore the number of shortcuts defined in the translations should match the number defined at the top level"
I'd perhaps say "would need to exactly match the number defined at the top level". Perhaps even say it could be treated as an error to not match the exact number and call this out as a drawback (eg. adding a new shortcut without adding a translation would break the translations)

phoglenix (approved):

PR #43, opened 2021-11-04
Attempted to generalize `ManifestOverride`

aarongustafson:
* Clarified what to do with excess array items
* Clarified what to do with disallowed keys

@mgiuca @loubrett Thoughts?

### DIFF
--- a/translations-explainer.md
+++ b/translations-explainer.md
@@ -4,7 +4,7 @@ Authors: [Aaron Gustafson](https://github.com/aarongustafson), Louise Brett (lou
 
 ## Overview
 
-This document proposes a new `translations` manifest member that enables web apps to provide alternate values for manifest fields for different languages. 
+This document proposes a new `translations` manifest member that enables web apps to provide alternate values for manifest members in specific languages. 
 
 ## Problem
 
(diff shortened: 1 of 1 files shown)


loubrett:
Thanks for adding this! Looks good to me.

I wonder how common it will be to localize shortcut icons. I think of these being universal icons.

aarongustafson:
> I wonder how common it will be to localize shortcut icons. I think of these being universal icons.

Agreed, but I think it probably makes sense to keep the door open to that possibility. After all, iconography is often culturally-tied, so it’s not outside the realm of possibility.

tomayac:
> > I wonder how common it will be to localize shortcut icons. I think of these being universal icons.
> 
> Agreed, but I think it probably makes sense to keep the door open to that possibility. After all, iconography is often culturally-tied, so it’s not outside the realm of possibility.

+1, the [Material Design Guide](https://material.io/design/usability/bidirectionality.html#localization) recommends flipping icons for RTL languages. One of the shortcut icons of Google Maps on iOS is "Send my location" with a Send icon (a paper plane flying to the right, which would fly to the left in RTL).

```html
<svg class="MuiSvgIcon-root MuiSvgIcon-fontSizeMedium MuiSvgIcon-root MuiSvgIcon-fontSizeLarge css-1shn170" focusable="false" viewBox="0 0 24 24" aria-hidden="true" data-testid="SendIcon" tabindex="-1" title="Send"><path d="M2.01 21 23 12 2.01 3 2 10l15 2-15 2z"></path></svg>
```

loubrett (approved):

kenchris:
What is up with this?

PR #47, opened 2021-12-06
Create user-preferences-explainer.md

loubrett:


user-preferences-explainer.md
```diff
@@ -0,0 +1,66 @@
...
+
+## Proposal
+
+Add a new dictionary manifest entry `user_preferences`, mapping preference strings to a [`ManifestOverride` object](#manifestoverride-object).
+
+Initially, the valid keys of the `user_preferences` member are `color_scheme_dark` and `color_scheme_light`. This could be expanded in the future to include other [user preference media features](#possible-extensions).
+
+### `ManifestOverride` Object
+
+The `ManifestOverride` is a generic object that contains a subset of redefined manifest properties appropriate to the context (e.g. `user_preferences`) in which the `ManifestOverride` is being used. The properties which may be redefined in the ManifestOverride object depends on the context (e.g. `user_preferences`). Any properties not allowed within the context will be ignored. The redefined fields override the values set in the root of the manifest.
```
tomayac:
```suggestion
The `ManifestOverride` is a generic object that contains a subset of redefined manifest properties appropriate to the context (e.g. `user_preferences`) in which the `ManifestOverride` is being used. The properties which may be redefined in the `ManifestOverride` object depend on the context (e.g. `user_preferences`). Any properties not allowed within the context will be ignored. The redefined fields override the values set in the root of the manifest.
```

tomayac (commented):
Should this explain why the explainer mentions [CSS](https://drafts.csswg.org/mediaqueries-5/#mf-user-preferences) and clearly is inspired by it, but doesn't use its syntax?

loubrett:
> Should this explain why the explainer mentions [CSS](https://drafts.csswg.org/mediaqueries-5/#mf-user-preferences) and clearly is inspired by it, but doesn't use its syntax?

Yes that's a good point, I've added bit in the overview section on this.

user-preferences-explainer.md
```diff
@@ -0,0 +1,68 @@
...
+
+## Proposal
+
+Add a new dictionary manifest entry `user_preferences`, mapping preference strings to a [`ManifestOverride` object](#manifestoverride-object).
+
+Initially, the valid keys of the `user_preferences` member are `color_scheme_dark` and `color_scheme_light`. This could be expanded in the future to include other [user preference media features](#possible-extensions).
+
+### `ManifestOverride` Object
+
+The `ManifestOverride` is a generic object that contains a subset of redefined manifest properties appropriate to the context (e.g. `user_preferences`) in which the `ManifestOverride` is being used. The properties which may be redefined in the `ManifestOverride` object depend on the context (e.g. `user_preferences`). Any properties not allowed within the context will be ignored. The redefined fields override the values set in the root of the manifest.
```
phoglenix: super-nit: "The properties **that** may be redefined"

user-preferences-explainer.md
```diff
@@ -0,0 +1,68 @@
...
+    },
+    "color_scheme_light": {
+        "theme_color": "#fff"
+        "background_color": "#fff"
+    }
+  }
+```
+
+When a user has dark mode enabled, the fields redefined for `color_scheme_dark` will be used instead of the top level fields.
+
```
phoglenix, line 51: You may also want an "alternatives considered" section to explain why you aren't using generic css queries as user preference selectors?

user-preferences-explainer.md
```diff
@@ -0,0 +1,68 @@
...
+*   Forced colors
+*   Prefers reduced data
+
+## Security and Privacy Considerations
+
+There are minimal security and privacy concerns with this proposal. This will allow sites to know what user preferences are set and use this for fingerprinting, however this information is already exposed through CSS.
+
+## Open questions
+
+How does this interact with `translations`  for fields that can be overridden by either `user_preferences` or `translations`, such as icons?
```
phoglenix, line 72: Perhaps add here a note about sites that use a site-specific "dark mode" theme?

phoglenix (approved):
LGTM, just some thoughts/nits

user-preferences-explainer.md
```diff
@@ -0,0 +1,68 @@
...
+
+Authors: [Aaron Gustafson](https://github.com/aarongustafson), Louise Brett (loubrett@google.com)
+
+## Overview
+
+Currently in the web app manifest a `theme_color` and `background_color` can be defined. However there is no way to change these based on dark mode or [other user preferences](#possible-extensions).
+
+This document proposes a new `user_preferences` manifest member that enables web apps to provide alternate values for manifest members given specific user preferences. The structure of this `user_preferences` field matches the structure of the proposed <code>[translations](https://github.com/WICG/manifest-incubations/blob/gh-pages/translations-explainer.md)</code> member.
+
+This new member is inspired by the [CSS media queries user preferences](https://drafts.csswg.org/mediaqueries-5/#mf-user-preferences). The keys are simple strings but they are derived from the CSS media query syntax and can easily be converted back if necessary.
```
tomayac: Converting back sets the expectation that things will then function as in CSS, which is what the issue discussion made clear would _not_ happen (see [this section](https://github.com/w3c/manifest/issues/975#:~:text=concerned%20about%20this.-,I%20don't%20think%20it's,go%20down%20this%20route.,-Contrary%20to%20what) specifically).

loubrett: Thanks for linking that, I had been looking at some older parts of the thread. I've removed that bit now.

user-preferences-explainer.md
```diff
@@ -63,6 +63,12 @@ In addition to the preferences `color_scheme_dark` and `color_scheme_light`, oth
 
 There are minimal security and privacy concerns with this proposal. This will allow sites to know what user preferences are set and use this for fingerprinting, however this information is already exposed through CSS.
 
+## Alternatives considered
+
+Use the CSS media query syntax for the keys (e.g. `(prefers-color-scheme: dark)`) and parse this as CSS. However, using the CSS parser adds a lot of complexity which we would like to avoid during manifest parsing.
```
phoglenix: You could phrase this a little more like an alternative, eg. "It would also be possible to use [...], instead of the fixed keys as proposed".
Maybe also expand on the complexity (which also happens during window opening and theoretically constantly), eg. "Before a web app window can launch, a CSS parser would need to be run to analyse the media query" and maybe give an example of a bad situation that is hard and not necessary, like a media query based on window size.

tomayac (approved):
Non-owner LGTM

phoglenix (approved):

PR #63, opened 2022-11-02
Rewrite web app launch procedures in terms of the web-app-launch spec algorithm

### LINKED ISSUE #67: "When launching a web app" feels handwavy
It seems like this is the entry point to the spec's algorithms, and it's not well defined when that happens.
This seems like a concept that should be defined in HTML, and which would then call this spec explicitly from there.

Such a concept may also be useful for https://github.com/w3c/navigation-timing/issues/180

/cc @domenic

alancutter:
Fixes: https://github.com/WICG/web-app-launch/issues/67

This updates 3 launch sites in manifest-incubations (note taking, protocol handlers and file handlers) to make use of the [=launch a web application with handling=] algorithm being added by: https://github.com/WICG/web-app-launch/pull/79/files

Note that the removal of `Launch queue and launch params` from this spec is due to these concepts already existing in https://wicg.github.io/web-app-launch/#script-interfaces-to-app-launches.

***
<a href="https://pr-preview.s3.amazonaws.com/alancutter/manifest-incubations/pull/63.html" title="Last updated on Nov 9, 2022, 2:51 AM UTC (c07bf86)">Preview</a> | <a href="https://pr-preview.s3.amazonaws.com/WICG/manifest-incubations/63/e63d465...alancutter:c07bf86.html" title="Last updated on Nov 9, 2022, 2:51 AM UTC (c07bf86)">Diff</a>

### DIFF
--- a/index.html
+++ b/index.html
@@ -23,7 +23,7 @@
             "dom",
             "file-system-access",
             "mimesniff",
-            "web-app-launch",
+            "web-app-launch"
             "window-controls-overlay"
           ],
           profile: "web-platform",
@@ -476,13 +476,13 @@
             <var>protocol_handler</var> having [=manifest=] <var>manifest</var>
             is invoked, it goes through the same steps used to [=invoke a
             protocol handler=] where the user agent, instead of [=navigating=]
-            to |resultURL|, SHOULD [=launch a web application=] passing
-            <var>manifest</var> and |resultURL|.
+            to <var>resultURL</var>, SHOULD [=launch a web application=]
+            passing <var>manifest</var> and <var>resultURL</var>.
           </p>
           <p class="issue">
             This should not invoke and alter [=invoke a protocol handler=] in
-            this way. The computation of |resultURL| should be extracted out
-            into a separate algorithm for general use.
+            this way. The computation of <var>resultURL</var> should be
+            extracted out into a separate algorithm for general use.
           </p>
         </section>
       </section>
@@ -860,7 +860,7 @@
                       a single element that is a {{FileSystemHandle}}
                       corresponding to |file|.
                       </li>
-                      <li>[=Launch a web application with handling=] passing
+                      <li><a data-cite="web-app-launch#dfn-launch-a-web-application-with-handling">Launch a web application with handling</a> passing
                       |manifest| and |params|.
                       </li>
                     </ol>
@@ -874,7 +874,7 @@
                   {{FileSystemHandle}}s that correspond to the file paths named
                   by |files|.
                   </li>
-                  <li>[=Launch a web application with handling=] passing
+                  <li><a data-cite="web-app-launch#dfn-launch-a-web-application-with-handling">Launch a web application with handling</a> passing
                   |manifest| and |params|.
                 </ol>
               </li>
--- a/index.html
+++ b/index.html
@@ -23,7 +23,6 @@
             "dom",
             "file-system-access",
             "mimesniff",
-            "web-app-launch"
             "window-controls-overlay"
           ],
           profile: "web-platform",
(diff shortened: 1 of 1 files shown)


index.html
```diff
@@ -476,9 +474,10 @@ <h3>
             When a <a>protocol handler description</a>
             <var>protocol_handler</var> having [=manifest=] <var>manifest</var>
             is invoked, it goes through the same steps used to [=invoke a
-            protocol handler=] defined in [=HTML=], where the user agent SHOULD
-            navigate to [=url=] and the appropriate browsing context is set to
-            a new top level browsing context.
+            protocol handler=] defined in [=HTML=] where the user agent,
```
marcoscaceres: It's usually best not to put which spec stuff if defined in as definitions get moved around to different docs: 

```suggestion
            protocol handler=] where the user agent,
```

alancutter: As a side this launch instruction seems quite iffy to me, I believe it intends to extract the |resultURL| variable in the algorithm referenced to know what URL to navigate to. I think it needs work but it's a bit out of scope for what I'm trying to do here.

index.html
```diff
@@ -281,11 +282,8 @@ <h3>
           <li>If the type of |manifest|["note_taking"]["new_note_url"] is not
           [=URL=], return.
           </li>
-          <li>Let |browsing context:Browsing Context| be the result of creating
-          a new [=top-level browsing context=].
-          </li>
-          <li>[=Navigate=] |browsing context| to resource
-          |manifest|["note_taking"]["new_note_url"].
+          <li>Run the steps to [=launch a web app with handling=] passing
```
mgiuca: Can you pass with explicit parameter names?

"setting _manifest_ to _manifest_ and _target URL_ to _manifest[note_taking][new_note_url]_".

index.html
```diff
@@ -476,9 +474,10 @@ <h3>
             When a <a>protocol handler description</a>
             <var>protocol_handler</var> having [=manifest=] <var>manifest</var>
             is invoked, it goes through the same steps used to [=invoke a
-            protocol handler=] defined in [=HTML=], where the user agent SHOULD
-            navigate to [=url=] and the appropriate browsing context is set to
-            a new top level browsing context.
+            protocol handler=] where the user agent,
```
mgiuca: Hmm this is messy.

Couple of thoughts:

1. Should we also hook into regular HTML protocol handlers so they trigger the launch as well? Or is this a special property you only get if you register using the app manifest? That's a big can of worms (you would need to edit HTML to call the launch with handler algorithm).
2. This existing text is probably wrong, or actually, redundant ([could be converted into non-normative](https://github.com/WICG/manifest-incubations/issues/64)). See above, the processing the manifest actually calls HTML's **register a protocol handler**. That means we should not need normative text to handle a protocol launch, it should just happen via HTML. So this text could just explain that that automatically happens. But that means we can't modify its behaviour to do anything different to plain HTML protocols.
3. Maybe we could just leave it hand-wavy for now.

mgiuca (approved):
lgtm

index.html
alancutter: I don't think we want manifest protocol handlers to ever not launch the web app, I wouldn't leave that detail up to user agents to work out.
This section is gnarly at the moment and I think reworking it is out of scope for this change so I opt for 3.

index.html
alancutter: Done.

index.html
```diff
@@ -851,25 +854,34 @@ <h2>
...
-                  by |files|.
+                  <li>Let |params:LaunchParams| be a new
+                  <a href="https://wicg.github.io/web-app-launch/#dom-launchparams">
+                  LaunchParams</a> with
+                  <a href="https://wicg.github.io/web-app-launch/#dom-launchparams-files">
+                  files</a> set to a {{FrozenArray}} of {{FileSystemHandle}}s
+                  that correspond to the file paths named by |files|.
                   </li>
-                  <li>[=launch a web app=] with |params|.
+                  <li><a href="https://wicg.github.io/web-app-launch/#dfn-launching-a-web-application-with-handling">
```
alancutter, line 882: These ugly href links are due to web-app-launch not existing in specref yet. I've filed https://github.com/tobie/specref/issues/721 to try and work out how to fix that.

PR #73, opened 2023-05-03
Add borderless mode explainer

sonkkeli:


borderless/display_override-borderless-explainer.md
```diff
@@ -0,0 +1,512 @@
+# Borderless Explainer
```
dmurph: Is there a reason you named this `display_override-borderless-explainer.md`? Can you make it `borderless-explainer.md` instead?

dmurph (commented):
One change requested :) otherwise it LGTM

borderless/display_override-borderless-explainer.md
sonkkeli: No reason, actually we can even just call it `explainer.md` as it's already in the borderless directory?

mgiuca: All the other explainers are in the top level and called `<feature-name>-borderless.md`. I think it's probably time to move these into an `explainers` subdir but for now, can you just put it in the top level?

borderless/explainer.md
```diff
@@ -0,0 +1,512 @@
...
+- Ensure draggable regions work with borderless mode
+
+## Non-goals
+
+- Implementation of
+  [additional windowing controls](https://github.com/ivansandrk/additional-windowing-controls/blob/main/awc-explainer.md),
+  which is another closely-related project
+- Making borderless mode dynamic (with e.g. JavaScript APIs or something else)
+- Changing display mode in app settings (e.g. from borderless to standalone)
+- Support for mobileOSs/macOS/windows
```
mgiuca: We shouldn't have a W3C spec that says supporting macOS and Windows are non-goals. This may not be a development priority for the team right now, but with my W3C hat on, we have to think about how this extends to other platforms as well as ChromeOS.

The explainer should explore whether this is at least possible on Windows, macOS and Linux (I believe it is).

(Note: I think it's OK to have a non-goal of supporting mobile, since there's very little distinction between `borderless` and `fullscreen` there.)

mgiuca:
Hi @sonkkeli 

Thanks for writing this up. I had a read over it. I will do a drive-by and give some in-line comments in a bit, but my overall thoughts is that the proposal makes sense, the write-up needs to be less Chrome and ChromeOS-focused.

Understanding that this is being primarily driven by ChromeOS, we still need to write an explainer that considers the API in the wider web ecosystem, given that it is being proposed as a W3C standard. I'll put direct comments in-line. I think a lot of this text could go into an internal design doc (public, if desired).

Note that it is still OK to reference specific browsers and OSes in explainers, but they need to be by way of example, not the exclusive focus of the document.

borderless/explainer.md
```diff
@@ -0,0 +1,512 @@
...
+and
+[display_override](https://github.com/WICG/display-override/blob/main/explainer.md)
+fields. Currently
+[supported values for display modes](<https://source.chromium.org/chromium/chromium/src/+/main:third_party/blink/public/mojom/manifest/display_mode.mojom?q=-file:(test%7CDebug%7C%5C.xtb$)%20case:yes%20third_party%2Fblink%2Fpublic%2Fmojom%2Fmanifest%2Fdisplay_mode.mojom&ss=chromium>)
+are `window-controls-overlay`, `tabbed`, `minimal-ui`, `standalone`, `browser`
+and `fullscreen`. See
+[Manifest/display_override#values](https://developer.mozilla.org/en-US/docs/Web/Manifest/display_override#values)
+for more. From the before-mentioned options, `window-controls-overlay` has some
+same qualities to what we are trying to achieve. However, it is still not
+offering enough native-like experience for some of the use-cases and thus this
```
mgiuca: I'm confused here. You say you need borderless for a "native-like experience", however, isn't the most "native-like experience", by definition, the `standalone` window mode? That allows the OS to draw its standard title bar on the window however it likes, making the window look exactly like native windows on the host OS.

WCO moves away from the native-like experience but still allows for the OS to draw its native control buttons.

Borderless moves completely away from the native-like experience, allowing the app to draw its own buttons (or not). Traditionally, borderless has been used for apps like [Steam](https://cdn.yurishwedoff.com/will_steam_ever_update_its_ui.jpg) to render a totally customized UI, the exact opposite of native.

I realize you are proposing this for a somewhat different use case, which is that a VDI app on a remote system will draw its own "native" UI, and we do not want the host OS to draw a second duplicate UI. However, let's be clear that, for example, on ChromeOS, the remote Windows UI is not "native"; it is foreign to the system, and that is exactly why you need a new API to draw it.

I think it would be good to break this into three distinct use cases as I see it, and explore each case and whether it is an explicit goal or not:

1. Apps that want to fully customize their title bar to still have the same functionality, but with their own style, like Steam.
2. Apps that want to completely remove the title bar and provide no controls (which can be seen sometimes in native apps for dialogs like splash screens).
3. The VDI case, where the remote app is drawing a remote-OS-native title bar and we want to avoid drawing a host-native title bar over the top of it or above it.

borderless/explainer.md
```diff
@@ -0,0 +1,512 @@
...
+explainer will explain how the title bar will be totally removed and enabling
+so-called borderless mode where the title bar area would be replaced with web
+content and so giving the developers full control on how the title bar would
+look like.
+
+## Example of a complex title bar with window-controls-overlay
+
+Some highly customized title bars would still look clumsy with
```
mgiuca: This is a weird example for me, because it looks like exactly the sort of case that WCO is designed for. The screenshot shows WCO awkwardly overlapping the other content, but the solution to this (in the general case) is not to remove the native controls, but to properly use WCO's `titlebar-area` region to draw the other stuff in the correct place.

However, I guess this screenshot is intended to be showing a remote app in a VDI context that we have no control over, which means you cannot use the `titlebar-area` to customize the layout. So you should explain this: the use case is not a "highly customized title bar", it's a remote title bar that we have no control over, which already provides its own control buttons.

borderless/explainer.md
```diff
@@ -0,0 +1,512 @@
...
+[VDI](https://www.softwaretestinghelp.com/best-vdi-software/#What_Is_Virtual_Desktop_Infrastructure)
+(Virtual Desktop Infrastructure) providers attempt to mirror remote applications
+to the local desktop. For example, the VDI provider's cloud is running a Text
+Processor (a desktop application window), which is to be presented on the client
+desktop as if the application was running locally.
+
+Installed web apps define their display mode in the manifest.json using display
+and
+[display_override](https://github.com/WICG/display-override/blob/main/explainer.md)
+fields. Currently
```
mgiuca: I would not enumerate all the values we currently support, since this might become out of date.

borderless/explainer.md
```diff
@@ -0,0 +1,512 @@
...
+{
+  "display_override": [ "borderless" ]
+}
+```
+
+### 2. Enabling draggable regions
+
+When the app enters the borderless mode,
+[draggable regions](https://github.com/WICG/window-controls-overlay/blob/main/explainer.md#defining-draggable-regions-in-web-content)
+`-webkit-app-region: drag;` will be enabled.
```
mgiuca: It's unfortunate that we have to rely on a webkit-prefixed property here, but I understand that is currently the case. Please link to https://github.com/w3c/csswg-drafts/issues/7017 to acknowledge that a standardized solution is coming, and that we would prefer developers switch to that when it is available.

borderless/explainer.md
```diff
@@ -0,0 +1,512 @@
...
+  }
+}
+```
+
+See:
+[https://developer.mozilla.org/en-US/docs/Web/CSS/@media/display-mode](https://developer.mozilla.org/en-US/docs/Web/CSS/@media/display-mode)
+
+### 4. Leveraging Window Management permission for borderless
+
+In order for the borderless mode to activate, the app needs to have Window
```
mgiuca: It makes me uncomfortable tying these two permissions together.

Borderless is a much scarier permission than management, due to the spoofing risk. I would not want every app I grant management permission to to magically get borderless permission as well.

mgiuca (changes requested):
Leaving a few comments. I have to run to a meeting now so I'll post these and continue the review later.

borderless/explainer.md
```diff
@@ -0,0 +1,512 @@
...
+set by an admin and later on edited in the app’s settings. For the **unmanaged**
+context, the Window management API permission must be granted.
+
+#### Enabling Window Management
+
+Granting the window management API permission can be done with the following
+JavaScript:
+
+```
+const screenDetails = await window.getScreenDetails();
```
mgiuca: I have a lot of reservations about how granting the permission (in the unmanaged case) works from an API and UI perspective.

1. There's a weirdness here about having to use JavaScript to ask for permission to use a feature declared in the manifest. This suggests further to me that it shouldn't be tied to the Window Management API permission.
2. What happens if the permission has not yet been granted or denied? (Presumably, we fall back to the next display mode in the chain, e.g. standalone., which means in the VDI case showing a double-title-bar.) Is this acceptable UX?
3. What happens at the instant the permission is granted? Do we expect the window's title bar to disappear upon being given permission? Or does it just change upon the next reload?
4. **My biggest concern**: If the window simply never showed the native title bar, it would be just a fact that this app doesn't have certain controls, and as you say below, there will be alternative ways to get to those features. However, having the app first appear with the native title bar, then a permission prompt being granted causing the title bar to disappear, leads the user to wonder where it went and how to get it back. I don't think it's good to provide UI (in the form of the permission prompt) to make it go away but no way to get it back (other than digging in app settings to revoke the permission). We should address this and possibly provide a way to "reveal" the hidden title bar.

If we're going to show the native title bar to the user on first run anyway, and the user will have to take action to make the native title bar go away, we could essentially make the "permission" a part of the UI instead of something you are prompted for. This is basically the same thought process we went through for WCO: if you're going to start with the full title bar shown, and the user will explicitly give permission to hide it, why not just make that permission a button in the UI, and why not make that button a toggle so they can get it back using the same mechanism they used before.

You could then show an animation when the user toggles the title bar away of the title bar collapsing into the system shelf (indicating where to go to find the controls), or perhaps just collapsing into a small clickable zone in the corner which lets you get it back.

None of this precludes having an admin policy to enable borderless by default. But in the unmanaged case (or managed case where the admin hasn't set that policy), it would be good to think about the UX of how the user grants and un-grants this, and it may make sense for it to not be a normal permission prompt.

borderless/explainer.md
```diff
@@ -0,0 +1,512 @@
...
+    console.log(status.state)
+  });
+```
+
+#### Disabling Window Management
+
+To disable borderless mode, the user can disable the Window Management
+permission in App Settings.
+
+In case the Window management permission is not provided or no longer in place,
```
mgiuca: Ah I see this answers my above question. I would rearrange this so you explain the flow up front: since in most cases you will start off with the permission not being provided, it would be good to explain first what the user sees when the permission has not been granted, and then how they get into this mode.

borderless/explainer.md
```diff
@@ -0,0 +1,512 @@
...
+the app will follow the normal display mode hierarchy based on the
+`display_override` and display values on the `manifest.json`. In case no other
```
mgiuca: If no other display mode is provided, the manifest spec requires you to assume `display: browser`. There are two possibilities here:

**First possibility: You actually mean to alter the fallback chain**

I suppose you could explicitly specify "if you have `display_override: borderless` then the default `display` is `standalone` but that seems like a weird edge case. I think it makes sense to just fall back to browser, and ask developers to ask for `standalone` if they want it.

**Second possibility: Borderless always behaves like standalone if not enabled**

Alternatively, you could explain this as not "falling back to standalone mode" but rather, "borderless mode will behave like standalone if the permission is not granted". There is a key difference here: you won't use the fallback chain at all. That is, if you write `display_override: borderless, browser`, then if the UA recognizes borderless, but the permission is not granted, it will still behave like standalone. Therefore, "In case no other display mode is provided" is not true -- it will default to standalone *regardless of whether any other display modes are provided*.

I think the answer depends on the above question, of how you intend the permission to be granted (if it dynamically switches between having a title bar and borderless, then the second possibility is the best).

borderless/explainer.md
```diff
@@ -0,0 +1,512 @@
...
+display mode is provided, it will
+[default to standalone mode](https://source.chromium.org/chromium/chromium/src/+/main:chrome/browser/web_applications/web_app_proto_utils.cc;drc=4a8573cb240df29b0e4d9820303538fb28e31d84;l=161)
```
mgiuca: Remove the link to Chrome code.

borderless/explainer.md
```diff
@@ -0,0 +1,512 @@
...
+
+### 5. Isolated Web App check
+
+Borderless mode is only available for
+[Isolated Web Apps (IWAs)](https://github.com/WICG/isolated-web-apps).
+
+## Related features
+
+- [app-details feature](crbug.com/1225871) (Googlers-only)
+- [go/cros-privacy-indicators-design](go/cros-privacy-indicators-design)
```
mgiuca: Remove Googlers-only links please. (These can go in a separate design doc, which can be public if you like.)

borderless/explainer.md
```diff
@@ -0,0 +1,512 @@
...
+  (Googlers-only)
+- [additional windowing controls](https://github.com/ivansandrk/additional-windowing-controls/blob/main/awc-explainer.md)
+
+### Displaying app’s origin in the app settings UI
+
+Since without the title bar and the 3-dot menu, the app’s origin won’t be visible
+anymore, it needs to be shown somewhere, in this case in App Settings. This
```
mgiuca: The general discussion here is good (about the security need to display the app's origin), but the explainer should not be specific about where it goes in ChromeOS UI, because this document is not a ChromeOS document.

"In this case in App Settings" and "This work has been completed" should be removed; instead, talk about the general problem and how a hypothetical user agent might solve it. You might say "On ChromeOS, for example, we intend to put this in the OS settings panel for the app", but only use it as an example, and do not assume the reader knows about OS-specific UI like "App Settings".

borderless/explainer.md
```diff
@@ -0,0 +1,512 @@
...
+work has been completed as part of the [app-details feature](crbug.com/1225871),
+but it is pending its launch.
+
+Displaying the origin is mainly necessary for debugging purposes. Until the app
```
mgiuca: This isn't true - the origin was a security requirement. That requirement may be changed with the introduction of IWAs, but we need to be explicit about what it was for in the first place.

borderless/explainer.md
```diff
@@ -0,0 +1,512 @@
...
+details get launched, an alternative way to get the origin of an IWA in
+borderless mode is via the developer console with
+[`document.location.href`](https://developer.mozilla.org/en-US/docs/Web/API/Location/href).
+
+In the future when app details have been launched and origin is available, the
```
mgiuca: This isn't really relevant; again, just talk about what a hypothetical OS might do in this case. Details about ChromeOS future plans can be moved to a separate doc.

borderless/explainer.md
sonkkeli: The permission design has been discussed and approved by security. I think the differentiation here is that it's only supported for isolated web apps, so providing the window management permission on its own would still not enable the feature.

borderless/explainer.md
```diff
@@ -0,0 +1,512 @@
...
+<span style="text-decoration:underline;">chrome://os-settings/app-management/detail?id=&lt;APP_ID></span>
+
+Mock of app settings UI with app’s origin:  
+![App settings UI mock](./images/app-settings-ui.png)
+
+### Privacy indicators
+
+Removing the title bar also removes the area where critical privacy indicators
+would live, e.g. for camera and microphone access. However there is already
+another on-going project to move those privacy indicators to the shelf.
```
mgiuca: Again, the "on-going project to move those" is ChromeOS-specific. We should just talk about how the user agent needs to find some way to display privacy indicators, e.g. in the operating system shelf.

The screenshots can be kept as an illustration of how we plan to address these issues.

borderless/explainer.md
```diff
@@ -0,0 +1,512 @@
...
+
+![Example privacy indicator](./images/privacy-indicators-in-shelf.png)
+
+See (Googlers only)
+[go/cros-privacy-indicators-design](go/cros-privacy-indicators-design) for more.
+
+## Other notable details
+
+### “Settings and more” three-dot button
```
mgiuca: This isn't necessarily a three-dot button. Officially, the standards do not require any sort of app settings, though many implementations have some kind of app settings menu.

Note that other platforms (like Android) have already solved this by moving some of the settings into the OS notifications tray, so this is not unprecedented. You could cite that precedent here.

Just say "some implementations of PWAs, e.g. Chrome and Edge, show a three-dot menu button (which gives users access to ....)". And then say user agents in borderless mode will want to find some other place to put these, or developers need to understand that they won't have those options.

borderless/explainer.md
```diff
@@ -0,0 +1,512 @@
...
+
+Since the three-dot button (which gives users access to extensions, security
+information about the page, access to cookies, etc.) will be disappearing, the
+most critical information (app’s origin) will be displayed in the app settings.
+
+The menu behind the three-dots button contains the following:
```
mgiuca: "Here is an example of the menu items provided by Chrome"

borderless/explainer.md
```diff
@@ -0,0 +1,512 @@
...
+
+<img src="./images/three-dot-menu.png" alt="Three-dot menu" width="450">
+
+<table>
+  <tr>
+   <td><strong>Button</strong></td>
+   <td><strong>Status</strong></td>
+  </tr>
+  <tr>
+   <td>App info</td>
```
mgiuca: I think this table is going into too much detail about specifics for one OS, and can be moved into a separate design doc, rather than the W3C explainer.

You could just generally say that user agents will need to drop or find a home for these sort of items that would normally appear on the app menu.

borderless/explainer.md
```diff
@@ -0,0 +1,512 @@
...
+  </tr>
+</table>
+
+A way to get the three-dots menu working would be that it could be part of the
+AWC feature, but that discussion should rather be part of the AWC feature. This
+could also be something considered to implement later on if seen necessary.
+
+## Demo
+
+Flags to enable:
```
mgiuca: Instructions for how to get the demo working with Chrome flags should probably be in a ChromeOS-specific document, and not here.

I think it's appropriate to keep links to your demo sites though, to illustrate the API.

borderless/explainer.md
```diff
@@ -0,0 +1,512 @@
...
+### Opening a popup from a borderless PWA
+
+- Opening a popup to any other origin → The popup should NOT be in borderless
+  mode (security risk if the following pop-up app would be malicious).
+- Opening a popup to itself (same origin) → The popup opens in borderless mode,
+  because the app has already got the permission to run in borderless mode.
+
+### Iframes
+
+- Borderless mode should not be available for iframes embedded inside of an IWA.
```
mgiuca: I'm not sure what this could mean (since the display mode is only available for a top-level browsing context, there is no possibility of borderless mode, or any display mode, being available for iframes).

I guess you could just reword it to say "Display modes cannot be enabled on embedded pages", to make it clear this isn't a new requirement we need to add, it's just a fact.

mgiuca (changes requested):
Thanks, I've completed the review.

I think most of my comments are basically around generalizing this for an audience of other browser vendors and W3C folks to understand what this means as a standard, rather than simply how it fits into ChromeOS plans. If there isn't one already, I think having a separate doc for ChromeOS implementation details (like moving permissions to the shelf, app settings, etc) would be good, instead of explaining those details here in a doc that's intended for a wider audience.

Beyond just the scope of the doc, I also thought some more about the permission model and whether it makes sense to be tied to the AWC and JavaScript APIs, and the UI for enabling it.

borderless/explainer.md
sonkkeli: It is not supported to open an Isolated Web App in browser mode so that means it is also not possible to have a borderless app that would be opened in browser mode. The latter sounds about right, I reworded the chapter to be more clear now. :)

borderless/explainer.md
sonkkeli: I rewrote the intro now, PTAL :) Also please resolve this if you think it's addressed.

borderless/explainer.md
sonkkeli: Yeah your assumption is correct, I rewrote that part too to underline that it's a remote title bar where it looks funky.

borderless/explainer.md
sonkkeli: Replaced with a link to the values

borderless/explainer.md
sonkkeli: 1. This could be prompted on window onload.

2. Yeah it displays standalone until we have the permission.

3. It immeadiately updates the window from standalone to borderless without having to refresh manually. If we toggle the permission from the settings, then it shows the default "Reload" infobar (which is default behaviour for toggling any other permission as well).

4. _(and the rest)_ I think here we have trusted on the first version not being available to other but the trusted VDI partners who we'd trust to not do evil things. This is because IWAs in general are not available in the open web yet. This should be something to take into consideration in the future though if extending it for non-partners. Do you think that should be something to already discuss in the explainer despite it would probably be implemented much later?

borderless/explainer.md
sonkkeli: I decided to list these instead as the picture was looking rather clumsy and didn't really bring much additional value to the text.

borderless/explainer.md
sonkkeli: I will move these to the run instructions of the demo-app. :) Any case these should be removed when everything is actually launched.

sonkkeli:
Hey Matt and thanks for all the comments, they made the explainer much better I think. There's still a few unresolved comments with some open Qs but in general I think it is ready for another PTAL . :)

borderless/explainer.md
mgiuca: nit: Change the link to Steam to just `https://store.steampowered.com/` instead of the random screenshot I gave you (it's a random thing I found on Google images). Even though that doesn't have a screenshot directly in the page, it's an authoritative link which people can use to find images if they need to.

Otherwise good.

borderless-explainer.md
```diff
@@ -108,21 +110,21 @@ The solution proposed consists of the following parts:
 ### 1. Setting the `display_override` to borderless
 
 To provide the maximum addressable area for web content, the User Agent (UA)
-will create a frameless window removing all UA provided Chrome, leaving only
-resizing of the window from its borders. The removed window controls will be
-enabled using AWC and HTML/JavaScript/CSS (see non-goals and
+will create a frameless window removing all UA provided by the browser, leaving
```
mgiuca: nit: UA -> UI
Drop "by the browser" since the UA is the browser. (Otherwise this sentence is saying "the UA removes all UI provided by itself".)

borderless/explainer.md
mgiuca: 1. Is not entirely satisfying to me but I understand it would work mechanically. Having a permission prompt that doesn't actually give the app any new API capabilities, it just makes something you declared in your manifest start working, doesn't feel like a permission to me. It feels like a UI setting. (Again, that's how WCO was designed. We could have made you do a permission prompt to activate WCO but we designed it as a UI setting instead.)
2. OK.
3. OK.
4. I don't think that addresses the question. I asked about what if the user grants the permission (which causes the window UI to disappear) and then wants to get it back again. There's no "evil thing" happening here, it's just a natural thing for a user setting to be reversible. The fact that, upon granting the permission, you will be hiding the entire permission granting UI somewhere else, means it can be tricky for a user to try and undo that change.

These things can probably be sorted out after the explainer lands in the repo, so I won't block the review on it. But I do want to keep talking about what this UI looks like in the non-enterprise non-VDI case. We should have a design in mind that scales beyond a set of trusted partners.

borderless/explainer.md
mgiuca: I'm assuming your rewording is this part:

> The `display_override` value will be ignored on unsupported OSs and it will act as a `standalone` window. This is because Isolated Web Apps default are not supported to open in `browser` display mode.

This still doesn't seem quite right. Basically:

1. On unsupported OSes (or in other words, user agents that have never heard of `borderless`), the `borderless` value will be ignored, and it will fall back to whatever is next in the list, or to the old `display` fallback chain. That is non-negotiable, since we can't change the behaviour of browsers who aren't implementing this feature.
2. On UAs that do understand `borderless`, but where the borderless permission has not been granted, they will fall back to behaving as if `standalone` was in place (regardless of what appears next in the `display_override` list or the value of `display`).

Does that seem right?

So I think you can just replace the above quoted paragraph with:

> User agents that do not implement `borderless` will follow the normal display mode hierarchy based on the `display_override` and `display` values in the manifest.

borderless-explainer.md
```diff
@@ -344,7 +265,6 @@ find the IWA related implementation details from IWA related documents.
 ```
 {
   "name": "Borderless title bar example",
-  "display": "standalone",
   "display_override": ["borderless"]
```
mgiuca, line 269: I see you removed `display: standalone`, but that is still required. User agents that do not understand `borderless` need to know what to fall back to. (PWAs should always provide `display` because by definition all the values in `display_override` are optional for UAs to implement.)

mgiuca (approved):
Thanks for revising! I think the overall changes to use Chrome/OS as an "example" rather than specifically designing for that platform are right what I had in mind.

There are a few comments and changes requested, both on this review and in some inline comments I sent earlier. I'll approve this so we don't need to do another round of reviews, but please address comments.

Cheers

borderless/explainer.md
sonkkeli: I already raised this Q up to our PM so we'll keep working on this. I will leave these comments open here.

dmurph (approved):
LGTM

sonkkeli:
I don't have the "power" to merge this btw. Could someone else click the button?

PR #74, opened 2023-05-03
Add borderless mode spec

sonkkeli:
***
<a href="https://pr-preview.s3.amazonaws.com/sonkkeli/manifest-incubations/pull/74.html" title="Last updated on May 31, 2023, 2:00 PM UTC (6ddfaf3)">Preview</a> | <a href="https://pr-preview.s3.amazonaws.com/WICG/manifest-incubations/74/ab4319d...sonkkeli:6ddfaf3.html" title="Last updated on May 31, 2023, 2:00 PM UTC (6ddfaf3)">Diff</a>

### DIFF
--- a/index.html
+++ b/index.html
@@ -64,12 +64,12 @@
         The [=manifest/display_override=] member of the [=application manifest=]
         is a <a>sequence</a> of <a data-cite=
         "appmanifest#dfn-display-modes-list">display mode list values</a>
-        including extensions like [=display mode/window-controls-overlay=].
-        This member represents the developer's preferred fallback chain for
-        [=display mode=]s. This field overrides the [=manifest/display=] member.
-        If the user agent does not support any of the [=display mode=]s
-        specified here, then it falls back to considering the
-        [=manifest/display=] member. See <a data-cite=
+        including extensions like [=display mode/window-controls-overlay=] and
+        [=display mode/borderless=]. This member represents the developer's
+        preferred fallback chain for [=display mode=]s. This field overrides the
+        [=manifest/display=] member. If the user agent does not support any of
+        the [=display mode=]s specified here, then it falls back to considering
+        the [=manifest/display=] member. See <a data-cite=
         "appmanifest#dfn-process-the-display-member">processing the display
         members</a> for the algorithm steps.
       </p>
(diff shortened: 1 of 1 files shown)


dmurph:
Looking at window-controls-overlay - that is a separate repository (which was one of the options we talked about). So options here:
1. If we want to put the explainer & spec inside of the manifest-incubations, which seems fine, I would recommend putting the spec text in the existing manifest-incubations spec `index.html`, which defines the display mode, etc.
2. If you want to keep the spec document separate, then in that case I would recommend creating a WICG repository exclusively for that display mode definition specification.

Unfortunately I'm not sure if having a nested spec html in this repo will work with the WICG infrastructure.

I don't think I have any major preference. @mgiuca do you?

mgiuca:
Hi @sonkkeli 

I think it doesn't matter _too_ much, but I agree with Dan it should either be in the manifest-incubations `index.html`, or a separate repo (either way, that would allow previews and such to work properly, which they don't if you add a new file in this repo).

My preference would be to just have it here. I would also prefer if WCO moved here as well. The fewer separate spec documents the better IMO, to avoid the need for weird cross-referencing and concepts split across many documents. If something is very separable and experimental then it can go in its own thing, but if there are cross-referenced concepts between WCO and this, then I would prefer them to both be here.

sonkkeli:
That sounds good to me. Meanwhile the WCO is in it's own repo, is it okay if I just reference e.g. the draggable regions etc. parts from there with a TODO and whenever that would be moved here, then those few references can be fixed?

sonkkeli:
I made some edits on this now so that there's no separate borderless spec but just added the borderless things to the manifest-incubations.

index.html
```diff
@@ -101,6 +102,74 @@ <h2>
...
+          <li>
+            <dfn data-lt="IWA">Isolated Web App</dfn>: feature for installed
+            and signed web applications which are packaged as Web Bundles and
+            can be used offline instead of being fetched over HTTPS.
+            More: https://github.com/WICG/isolated-web-apps
+          </li>
+          <li>
+            <dfn data-lt="draggable region">Draggable regions</dfn>. To define
+            which regions or elements in the title bar are draggable, this
+            specification defines the CSS <code><dfn>app-region</dfn></code>
```
dmurph: my worry here is that we don't support `app-region` yet, it's currently only supported as `-webkit-app-region`. Should we mention this?

See the note in the web.dev article:
https://web.dev/window-controls-overlay/#determining-which-parts-of-the-title-bar-are-draggable

My guess is that we should mention something like this note? I'm not sure the best wording here.

index.html
```diff
@@ -101,6 +102,74 @@ <h2>
...
+          [=manifest/display_override=] also supports certain extensions to it.
+        </p>
+        
+        <dt>
+          <dfn data-export="" data-dfn-for="display mode">
+            borderless
+          </dfn>
+        </dt>
+        <dd>
+          The web application opens without any host-native title bar or
```
dmurph: The web application does not have any host-native title bar or [=window controls=] visible and with the web contents extended to the whole title bar area. The app can specify draggable regions in the web contents to create a customized title bar. The user agent may change the title bar state depending on various security considerations, like an out-of-scope navigation.

index.html
```diff
@@ -101,6 +102,74 @@ <h2>
...
+          [=window controls=] visible and with the web contents extended to the
+          whole title bar area. The app will also be able to specify draggable
+          regions in the web contents to create a customized title bar.
+          Borderless mode defaults to `standalone` if it's used in a
+          non-Isolated Web App (IWA) or without Window Management permission.
```
dmurph: This shouldn't be specified like this - if the user agent doesn't support 'borderless', then it goes to the next display mode in the `display_override` list.

(e.g. If that is `standalone`, then great. Otherwise it defaults to using `display`, and if that is empty, it defaults to `browser`)

index.html
```diff
@@ -101,6 +102,74 @@ <h2>
...
+        </dd>
+
+        <dt>
+          <dfn data-export="" data-dfn-for="display mode">
+            window-controls-overlay
+          </dfn>
+        </dt>
+        <dd>
+          The web application opens with native OS window controls visible but
```
dmurph: The display mode describes the state, not the opening behavior - maybe reword a bit?

index.html
sonkkeli: I added a second paragraph regarding the standardization. WDYT?

sonkkeli:
Hey @dmurph PTAL :)

index.html
```diff
@@ -101,6 +101,78 @@ <h2>
...
+
+        <dt>
+          <dfn data-export="" data-dfn-for="display mode">
+            window-controls-overlay
+          </dfn>
+        </dt>
+        <dd>
+          The web application has native OS window controls visible but the web
+          contents is extended to the rest of the title bar area. The app can
+          specify draggable regions in the web contents to create a customized
```
dmurph: link to the dfn you created above

(I think `[=draggable region=]`?)

index.html
```diff
@@ -101,6 +101,78 @@ <h2>
...
+        
+        <dt>
+          <dfn data-export="" data-dfn-for="display mode">
+            borderless
+          </dfn>
+        </dt>
+        <dd>
+          The web application does not have any host-native title bar or
+          [=window controls=] visible and with the web contents extended to the
+          whole title bar area. The app can specify draggable regions in the web
```
dmurph: link to the dfn you created above

(I think [=draggable region=]?)

index.html
```diff
@@ -101,6 +101,78 @@ <h2>
...
+          contents to create a customized title bar. The user agent may change
+          the title bar state depending on various security considerations, like
+          an out-of-scope navigation.
+        </dd>
+
+        <dt>
+          <dfn data-export="" data-dfn-for="display mode">
+            window-controls-overlay
```
dmurph: My only worry here is that we already have a definition here: https://wicg.github.io/window-controls-overlay/#dfn-window-controls-overlay-0. And this document already correctly links there.

Without migrating the window controls overlay spec into this one, I think we should

- If possible, just link to common concepts in the window-controls-overlay spec. This might require you to export them from that spec.
- (if not, it's fine to duplicate them)

### Option 1: Keep the 'extensions' for display_override section

- Have this section link to window-controls-overlay, instead of defining it.
- Keep 'borderless' definition in here.
- Link to this section specifically in the "display_override member" section as extensions, instead of listing them again in the txt.

### Option 2: Change the 'extension' section to just 'borderless' display mode.

- Have this section just be describing the 'borderless' display mode (so it isn't designed to encompass all extensions) and remove the window-controls-overlay definition.

What do you think?

@mgiuca interested in your thoughts if you have cycles.

mgiuca: IMO we should not have cyclic dependencies between these specs. I see these documents as "staging" drafts before they get merged upstream into the "parent" doc. The dependencies should be like this:

WCO -> Manifest Incubations -> Manifest

If Manifest Incubations has a cyclic dependency on WCO, then it implies the concepts in WCO are mature enough to be promoted into MI.

Ultimately, WCO should be entirely folded into MI. But to avoid biting off the entire burger at once, could we just migrate any concepts we need right now into MI? Thus I propose Option 3:

- Copy the common concepts from the WCO spec to MI. (Are we just talking about draggable regions, or something more?)
- Submit a pull request to WCO to remove those concepts and link to MI instead.

sonkkeli: SGTM! And yes, draggable regions is the one we are interested about with borderless.

Here would be the corresponding changes to the WCO spec:
https://github.com/WICG/window-controls-overlay/compare/main...sonkkeli:window-controls-overlay:move-draggable-regions-to-manifest-incubations-spec?expand=1

I'll create a PR once this one is merged?

sonkkeli:
PTAL :)

index.html
```diff
@@ -101,35 +101,115 @@ <h2>
...
+        <h3>
+          Defining <dfn data-export="" data-lt="draggable region">draggable
+            regions</dfn>
+        </h3>
+        <p>
+          To define which regions or elements in for example a title bar are
+          draggable, this specification defines the CSS
+          <dfn>`app-region`</dfn> property.
+        </p>
+        <p>
```
mgiuca: This is a weird paragraph to have _in the standard_. We are defining `app-region` right here, so we shouldn't say right below that we're recommending `-webkit-app-region`.

I would say this should be turned into a non-normative note, but I think even that goes too far: we should not recommend webkit-prefixed things in a standard.

Perhaps as a compromise, we could turn this para into a `class="issue"` and say "At-risk feature: `app-region` has not been implemented in any user agent, so is at risk. Note that some user agents use the non-standard `-webkit-app-region` for the same purpose." (Note: This doesn't say WAR "shall be used as the supported"; it just points out the fact that it exists.)

index.html
mgiuca: That's good. Feel free to submit that other PR at the same time and point to the fact that it's being added to incubations. (I don't think it'll be controversial, given that it contains a paragraph that literally says it "might belong in other specifications".)

index.html
sonkkeli: PTAL if this'd be good:

image

dmurph: optional nit to do the definition separate from the issue block. (so the second P and points occur separatly, and the app-region is 'defined' there, and then the first P is kept in the issue block). But I'm also fine as-is, having the whole thing as an issue communicates this probably better.

dmurph (approved):
This LGTM

index.html
sonkkeli: Done

PR #95, opened 2024-03-06
Add tabbed mode spec

mgiuca:
Based on an earlier draft by @loubrett (Louise Brett).

***
<a href="https://pr-preview.s3.amazonaws.com/mgiuca/manifest-incubations/pull/95.html" title="Last updated on Apr 26, 2024, 5:52 AM UTC (57014a6)">Preview</a> | <a href="https://pr-preview.s3.amazonaws.com/WICG/manifest-incubations/95/db88c85...mgiuca:57014a6.html" title="Last updated on Apr 26, 2024, 5:52 AM UTC (57014a6)">Diff</a>

### DIFF
--- a/index.html
+++ b/index.html
@@ -282,31 +282,29 @@
           [[urlpattern]] for more information.
         </aside>
         <p>
-          If the [=home_tab=] member is validly specified, every application
-          window SHOULD feature an [=application context=] called the <dfn>home
-          tab context</dfn> that has special properties compared to other
-          application contexts. If [=home_tab=] is not present or not valid,
-          then the application windows SHOULD NOT have a [=home tab context=].
+          The <dfn>home tab context</dfn> is an optional [=application
+          context=] that has special properties compared to other application
+          contexts. If the [=home_tab=] member is validly specified, every
+          application window SHOULD feature a [=home tab context=]. If
+          [=home_tab=] is not present or not valid, then the application
+          windows SHOULD NOT have a [=home tab context=].
         </p>
         <p>
           How the [=home tab context=] is presented is at the discretion of the
           user agent, but it SHOULD have a different appearance to normal tabs.
         </p>
         <p>
-          The <dfn>home tab scope</dfn> is a set of [=URLs=] which comprises
-          the union of the [=start URL=] and all URLs |url:URL| to which
-          applying [=URL pattern/match=] given any element of the
-          [=scope_patterns=] field and |url| returns a {{URLPatternResult}}.
+          The <dfn>home tab scope</dfn> is a set of [=URLs=] which, if
+          [=home_tab=] is validly specified, comprises the union of the [=start
+          URL=] and all URLs |url:URL| to which applying [=URL pattern/match=]
+          given any element of the [=scope_patterns=] field and |url| returns a
+          {{URLPatternResult}}. If [=home_tab=] is not present or not valid,
+          then the [=home tab scope=] is the empty set.
         </p>
         <p>
           A URL is said to be <dfn>within home tab scope</dfn> if it is a
           member of the [=home tab scope=]. It is <dfn>outside of home tab
           scope</dfn> if it is not a member of [=home tab scope=].
-        </p>
-        <p class="issue">
-          Be more clear that [=home tab scope=] is not defined when `home_tab`
-          is not given (as opposed to the above, which implies that it defaults
-          to [=start URL=]).
         </p>
         <aside class="note">
           The [=home tab scope=] is the set of URLs to which the [=home tab
--- a/index.html
+++ b/index.html
@@ -262,71 +262,112 @@
         </h3>
         <p>
           The `home_tab` member of the [=tab_strip=] object is an ordered map
-          that contains information about a specially marked [=application
-          context=] that, if it exists, is in every app window. This
-          [=application context=] has special properties compared to other
-          application contexts, such as restricted navigation and it SHOULD
-          have a different appearance.
-        </p>
-        <p>
-          A <dfn>home tab context</dfn> is the [=application context=] in which
-          these special properties are applied.
-        </p>
-        <aside class="note">
-          The home tab is intended to serve as a top level menu for the app.
-        </aside>
-        <p>
-          If the [=tab_strip/home_tab=] member is not present, then the
-          application is not intended to have a [=home tab context=], otherwise
-          it is an ordered map with the following members:
+          that contains information about a special "home tab" that is intended
+          to serve as the top-level menu for the application. It contains the
+          following members:
         </p>
         <ul>
           <li>[=home_tab/scope_patterns=]
           </li>
         </ul>
         <p>
-          How the [=home tab context=] is presented is at the discretion of the
-          user agent.
-        </p>
-        <p>
-          The <dfn>home tab scope</dfn> is the set of URLs to which the home
-          tab context can be navigated to. All navigations within the
-          application to a URL within the [=home tab scope=] will be done in
-          the [=home tab context=], and all other navigations will be done in a
-          different [=application context=].
-        </p>
-        <p>
-          The [=home tab scope=] consists of the [=start URL=] and URLs that
-          match a pattern specified in the [=scope_patterns=] field.
+          The <code><dfn>scope_patterns</dfn></code> member is a list of
+          {{URLPatternInput}}s that define the [=home tab scope|scope of the
+          home tab=] relative to the [=manifest URL=].
         </p>
         <aside class="note">
-          Navigation to a URL within the [=home tab scope=] will open in the
-          home tab, leaving behind the tab that initiated the navigation.
-          Developers should ensure the tab left behind does not end up in an
-          unexpected state.
+          Each scope pattern is either a string which looks like a URL but has
+          "<code>*</code>" as a wildcard character, or a dictionary containing
+          separate URL components with "<code>*</code>" wildcards. See
+          [[urlpattern]] for more information.
         </aside>
+        <p>
+          If the [=home_tab=] member is validly specified, every application
+          window SHOULD feature an [=application context=] called the <dfn>home
+          tab context</dfn> that has special properties compared to other
+          application contexts. If [=home_tab=] is not present or not valid,
+          then the application windows SHOULD NOT have a [=home tab context=].
+        </p>
+        <p>
+          How the [=home tab context=] is presented is at the discretion of the
+          user agent, but it SHOULD have a different appearance to normal tabs.
+        </p>
+        <p>
+          The <dfn>home tab scope</dfn> is a set of [=URLs=] which comprises
+          the union of the [=start URL=] and all URLs |url:URL| to which
+          applying [=URL pattern/match=] given any element of the
+          [=scope_patterns=] field and |url| returns a {{URLPatternResult}}.
+        </p>
+        <p>
+          A URL is said to be <dfn>within home tab scope</dfn> if it is a
+          member of the [=home tab scope=]. It is <dfn>outside of home tab
+          scope</dfn> if it is not a member of [=home tab scope=].
+        </p>
+        <p class="issue">
+          Be more clear that [=home tab scope=] is not defined when `home_tab`
+          is not given (as opposed to the above, which implies that it defaults
+          to [=start URL=]).
+        </p>
+        <aside class="note">
+          The [=home tab scope=] is the set of URLs to which the [=home tab
+          context=] can be navigated. All navigations within the application to
+          a URL [=within home tab scope=] will be performed in the [=home tab
+          context=], and all navigations to a URL [=outside of home tab scope=]
+          will be performed in a different [=application context=], creating
+          one if necessary. The application's [=start URL=] is always part of
+          the [=home tab scope=].
+        </aside>
+        <p>
+          When a [=home tab scope=] is defined, the user agent SHOULD ensure
+          that at all times:
+        </p>
+        <ul>
+          <li>every application window has exactly one [=home tab context=],
+          and
+          </li>
+          <li>every [=home tab context=]'s [=browsing context/initial URL=] is
+          [=within home tab scope=], and
+          </li>
+          <li>every non-home-tab [=application context=]'s [=browsing
+          context/initial URL=] is [=outside of home tab scope=].
+          </li>
+        </ul>
+        <p class="issue" title="TODO: Verify understanding">
+          The rule about "[=initial URL=] is [=within home tab scope=]" is
+          based on the understanding that "initial URL" refers to the current
+          URL of the document (not simply the URL initially loaded into the
+          application context). The "initial" simply refers to not taking into
+          account changes made to the display URL via the History API. Verify
+          this and possibly add a note. Also, this is not linkable. Find a way
+          to reference this aspect of the context (possibly via document).
+        </p>
+        <p>
+          In order to ensure the above rule that "every application window has
+          exactly one [=home tab context=]", the user agent will need to create
+          a new [=home tab context=] whenever a new application window is
+          created, for example when launching the application, or when moving a
+          tab to a new window. A newly created [=home tab context=] SHOULD be
+          navigated to the [=start URL=], which by definition is [=within home
+          tab scope=].
+        </p>
         <section>
-          <h4>
-            <code><dfn>scope_patterns</dfn></code> member
-          </h4>
+          <h3>
+            Navigations concerning the home tab scope
+          </h3>
           <p>
-            The [=home_tab/scope_patterns=] member is a list of [=URL
-            pattern=]s that define the scope of the home tab.
-          </p>
-          <p>
-            If navigating from the [=home tab context=] to a [=URL=] |URL:url|
-            that is outside of the [=home tab scope=], the following steps are
-            run:
+            If navigating from the [=home tab context=] to a [=URL=] |url:URL|
+            that is [=outside of home tab scope=], the following steps are run:
           </p>
           <ol>
-            <li>Create a new [=application context=] in the current window.
+            <li>Create a new [=application context=] in the current window and
+            focus it.
             </li>
             <li>Navigate it to |url|.
             </li>
           </ol>
           <p>
-            If navigating to a [=URL=] |URL:url| within the [=home tab scope=],
-            the following steps are run:
+            If navigating to a [=URL=] |url:URL| [=within home tab scope=], the
+            following steps are run:
           </p>
           <ol>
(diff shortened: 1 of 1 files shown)


mgiuca:
@loubrett Ready for you to look now.

index.html
```diff
@@ -84,76 +84,87 @@ <h2>
...
+            considerations, like an out-of-scope navigation.
+          </dd>
+          <dt>
+            [=display mode/window-controls-overlay=]
+          </dt>
+          <dt>
+            <dfn data-export="" data-dfn-for="display mode">tabbed</dfn>
+          </dt>
+          <dd>
+            The web application can have multiple application contexts combined
```
loubrett: Should 'application context' here be a link? Same for the other one in this paragragh and more at line 251 and 252.

index.html
```diff
@@ -186,31 +197,272 @@ <h4>
...
+        <p>
+          A <dfn>home tab context</dfn> is the [=application context=] in which
+          these special properties are applied.
+        </p>
+        <aside class="note">
+          The home tab is intended to serve as a top level menu for the app.
+        </aside>
+        <p>
+          If the [=tab_strip/home_tab=] member is not present, then the
+          application is not intended to have a home tab context, otherwise it
```
loubrett: Should 'home tab context' here be a link? Same for line 301

index.html
```diff
@@ -186,31 +197,272 @@ <h4>
...
+          </li>
+        </ul>
+        <p>
+          How the [=home tab context=] is presented is at the discretion of the
+          user agent.
+        </p>
+        <p>
+          The <dfn>home tab scope</dfn> is the set of URLs to which the home
+          tab context can be navigated to. All navigations within the
+          application to a URL within the home tab scope will be done in the
```
loubrett: Should 'home tab scope' here be a link?

index.html
```diff
@@ -186,31 +197,272 @@ <h4>
...
+          when clicked/activated, opens a new [=application context=] within
+          the application window. It has the following members:
+        </p>
+        <ul>
+          <li>[=new_tab_button/url=]
+          </li>
+        </ul>
+        <p>
+          If [=new_tab_button/url=] is within the [=home tab scope=], then the
+          new tab button may be hidden.
```
loubrett: Should 'may' be capitalized?

loubrett (commented):
Just some minor comments about linking stuff.

I haven't thought through the URLPattern stuff any more yet so I'll have a think about that next week.

index.html
mgiuca: Done. I won't repeatedly link multiple times per paragraph.

index.html
mgiuca: Done.

index.html
mgiuca: Done.

index.html
mgiuca: Good call-out. I think it's actually good to strengthen this to a MUST.

And further, I think this rule should be moved to the parser: instead of simply saying "don't show the button in the UI", the parser should just ignore it if it's within scope. Would that change the behaviour at all? (I don't think so, although internally it would mean we are discarding the illegitimate new_tab URL rather than storing it in the app database, which is slightly risky if we want to change this rule later.)

loubrett: Yeh we could change the parser to ignore it.
Currently later in the code, if the new tab URL is unset, we use the start_url (which will then always be in the home tab scope so it gets hidden). But we could leave it unset, and make sure the parser sets it to start_url when relevant.
I don't know how likely it is to want to change this rule, seems like very strange behaviour if we allowed it.

mgiuca:
I just pushed a major rework. Still incomplete so don't re-review yet.

mgiuca:
Hi @loubrett . This is finally ready for review. PTAL; note that you can preview it [here](https://pr-preview.s3.amazonaws.com/mgiuca/manifest-incubations/pull/95.html).

mgiuca:
@dmurph adding as editor. PTAL, thanks, and let me know if you want to discuss!

index.html
```diff
@@ -218,6 +237,400 @@ <h2>
...
+        <p>
+          The <dfn>home tab context</dfn> is an optional [=application
+          context=] that has special properties compared to other application
+          contexts. If the application [=has a home tab=], every application
+          window SHOULD feature a [=home tab context=]. If not, then the
+          application windows SHOULD NOT have a [=home tab context=].
+        </p>
+        <p>
+          How the [=home tab context=] is presented is at the discretion of the
+          user agent, but it SHOULD have a different appearance to normal tabs.
```
loubrett: Should this say `normal application contexts` instead of `normal tabs`? You don't really use the word tab anywhere else.

index.html
```diff
@@ -218,6 +237,400 @@ <h2>
...
+        <aside class="note">
+          Each scope pattern is either a string which looks like a URL but has
+          "<code>*</code>" as a wildcard character, or a dictionary containing
+          separate URL components with "<code>*</code>" wildcards. See
+          [[urlpattern]] for more information.
+        </aside>
+        <p>
+          An application <dfn>has a home tab</dfn> if the [=Document/processed
+          manifest=] includes a non-null [=home_tab=] member of the
+          [=tab_strip=] member.
```
loubrett: Should the `has a home tab` definition also  say the display mode needs to be tabbed? Or is it implied that none of these fields are relevant in that case?

index.html
```diff
@@ -218,6 +237,400 @@ <h2>
...
+        <aside class="note">
+          <p>
+            The "home tab scope" (i.e., the set of [=URLs=] which are [=within
+            home tab scope=]) is the set of URLs to which the [=home tab
+            context=] can be navigated . All navigations within the application
+            to a URL [=within home tab scope=] will be performed in the [=home
+            tab context=], and all navigations to a URL [=outside of home tab
+            scope=] will be performed in a different [=application context=]
+            (i.e., another tab), creating one if necessary. The application's
+            [=start URL=] is always part of the home tab scope.
```
loubrett: Nit (and probably doesn't matter), but here you say the start URL is always in the home tab scope but then below you give a case where it isn't (because the scope is the empty set).

index.html
```diff
@@ -218,6 +237,400 @@ <h2>
...
+            window is created (for example when launching the application, or
+            when moving a tab to a new window), the user agent MUST create a
+            new [=home tab context=] in that window. A newly created [=home tab
+            context=] SHOULD be navigated to the [=start URL=], which by
+            definition is [=within home tab scope=].
+          </p>
+        </section>
+        <section>
+          <h3>
+            Navigations concerning the home tab scope
```
dmurph, line 364: As per my exploration for link capturing, this has potential to break sites. I recommend linking the 'navigating' keyword here into somewhere in the html spec. Likely in 

Specifically, I recommend probably saying something along the lines of "When [creating a new top-level browsing context and document](https://html.spec.whatwg.org/multipage/document-sequences.html#creating-a-new-top-level-browsing-context),  or [creating a new auxilary browsing context and document](https://html.spec.whatwg.org/multipage/document-sequences.html#creating-a-new-auxiliary-browsing-context), choose the 'home' tab's browsing context when [creating a new browsing context](https://html.spec.whatwg.org/multipage/document-sequences.html#creating-a-new-browsing-context)." (that's actually probably not great to add the 'conditional execution of algo C if you are in algo A or B - perhaps we can just modify C w/o the condition? But this is probably enough to explain my request).

Or -perhaps you'll want to say something like "create a new browsing context that replaces the home tab browsing context".

Example of where this can cause error - if you have named frames, and you open a non-home-tab url (foo.com/away, named frame "B"), and then a user clicks a link that navigates the "B" frame to foo.com, which is in the home tab context.

Or - if the user clicks on a navigate-self link in a tab that is not the home tab, but navigates into the home tab context, what do you want?

It would be really helpful to reference the spec here around navigations, and this helps prevent things like breakages with window.opener and window.frames - the way this is implemented it has to be done where the browsing context gets chosen or created for the navigation in order to not break that.

index.html
```diff
@@ -218,6 +237,400 @@ <h2>
...
+            the home tab. Similarly, navigation from a regular tab to a URL
+            [=within home tab scope=] will open in the home tab, leaving the
+            existing document open in the tab that initiated the navigation.
+            Developers should ensure that in both of these cases, the document
+            left behind does not end up in an unexpected state.
+          </aside>
+        </section>
+        <section class="informative">
+          <h3>
+            Home tab invariants
```
dmurph, line 419: These invariants are likely 'cleaned up' or explained better by my above suggestion about exactly where the 'home tab context' is chosen to be used for a navigation. Then this is more of an informative note / explanation of the behavior, and the edge cases are:
- window.opener (actually this might not be a huge deal - you can just choose the home browsing context for the navigation & allow that to be part of the browsing group the link came from)
- frame name is specified, which already exists
- self navigation (kinda the same as frame name)
- the history modification API as you mentioned
- others?

index.html
```diff
@@ -218,6 +237,400 @@ <h2>
...
+        <aside class="note">
+          The [=new_tab_button=] member will be ignored if it is [=within home
+          tab scope=].
+        </aside>
+        <p>
+          When the new tab button is activated by the end user, the following
+          steps are run:
+        </p>
+        <ol>
+          <li>Create a new [=application context=] in the current window and
```
dmurph: perhaps reference the HTML spec algo too somehow? https://html.spec.whatwg.org/multipage/document-sequences.html#creating-a-new-browsing-context

index.html
```diff
@@ -218,6 +237,400 @@ <h2>
...
+          |manifest|["start_url"].
+          </li>
+        </ol>
+        <section>
+          <h3>
+            Processing the `home_tab` member
+          </h3>
+          <p>
+            To <dfn>process the `home_tab` member</dfn>, given [=ordered map=]
+            |json:ordered map|, [=ordered map=] |manifest tab strip:ordered
```
dmurph: perhaps name this `json_tab_strip`  to help readability?

index.html
```diff
@@ -218,6 +237,400 @@ <h2>
...
+            would open the page in the home tab.
+          </aside>
+        </section>
+        <section>
+          <h3>
+            Processing the `scope_patterns` member
+          </h3>
+          <p>
+            To <dfn>process the `scope_patterns` member</dfn>, given [=ordered
+            map=] |json:ordered map|, [=ordered map=] |manifest home
```
dmurph: same? where are we in the json of the manifest?

index.html
dmurph: maybe rename this here and elsewhere `manifest_tab_strip_member`?

Or- not sure what is 'normal' for spec - just wanted to make it clearer this was a variable that is in the parsed manifest output. Fine to keep as-is if it matches convension.

index.html
```diff
@@ -218,6 +237,400 @@ <h2>
...
+            exactly.
+          </p>
+          <p>
+            If the application does not [=has a home tab|have a home tab=],
+            then the home tab scope is the empty set.
+          </p>
+        </aside>
+        <section>
+          <h3>
+            Every window has a home tab
```
dmurph, line 351: There are some assumptions in this text that the browsing contxt is in `tabbed` display mode. The definition of the `application context` doesn't differentiate windowed vs in the browser. It might be good to clarify that the context for some of these assertions is that the window is in tabbed display mode.... you may need to define a mulit-browser-context window? Not sure if that is in spec somewhere.

dmurph (commented):
Some initial comments - also happy to go over in-person in a meeting

index.html
mgiuca: Done.

index.html
mgiuca: Hmm, yes, I think you're right. It's a little tricky since technically applications don't have display modes, application contexts do. So it's a little wrong to say "if the application's display mode is X" but I think we just need to go with that since that clearly shows the intention.

We do need something like this because later we say "whenever a new application window is created ... the user agent MUST create a new home tab context" which clearly can't be true for non-tabbed apps.

index.html
mgiuca: "and probably doesn't matter" --> it's a web standard, everything matters! :joy:

index.html
mgiuca: This section is intended to be informative. I intend for this to be (non-normatively) the overriding philosophy of how tabs work, while the normative text makes sure that it happens.

I'll have to think about window.opener and frames which I hadn't previously considered. However I think frames don't matter because this only applies to top-level browsing contexts.

index.html
mgiuca: Actually there is an existing definition of [create a new application context](https://www.w3.org/TR/appmanifest/#dfn-create-a-new-application-context) which I can leverage here.

index.html
mgiuca: Louise just raised this: I have amended the definition of "has a home tab" to require that the display mode is tabbed, which I think addresses this.

> The definition of the application context doesn't differentiate windowed vs in the browser.

I'm pretty sure it does (though not very clearly) - the definition of "[application context](https://www.w3.org/TR/appmanifest/#dfn-application-context)" is "A top-level browsing context that has a manifest applied to it" which my understanding means an application window. Otherwise I think a lot of our assumptions throughout the Manifest spec are flawed.

index.html
mgiuca: I'll need to block out some more time to think about this... starting off, adding a link to "navigate".

I think it's OK to just hand-wavily say "if navigating from...", after all that's roughly what the definition of NavigateEvent does ([link](https://html.spec.whatwg.org/multipage/nav-history-apis.html#the-navigate-event)).

dmurph: If you can, I think it would be best to hook this into the navigation algorithm. I think that should be possible - I think the correct behavior here is to say that when a navigation occurs within an application context that is 'tabbed', if that navigation [creates a new browsing context](https://html.spec.whatwg.org/multipage/document-sequences.html#creating-a-new-browsing-context), then it should use the 'home tab' browsing context.

What do you think?

index.html
mgiuca: We had an offline discussion about this. The summary is:

1. The paragraph that explains to developers of single-page applications that they should do real navigations in and out of home scope, is to be reworded so it's clear that it's giving developers advice, rather than highlighting an issue.
2. The "navigations concerning the home tab scope" should be reworked (and we workshopped a sketch of how the algorithm should look, which is as follows):

> When [choosing a navigable for a navigation ](https://html.spec.whatwg.org/multipage/document-sequences.html#the-rules-for-choosing-a-navigable) from the [home tab context](https://pr-preview.s3.amazonaws.com/mgiuca/manifest-incubations/pull/95.html#dfn-home-tab-context) to a [URL](https://url.spec.whatwg.org/#concept-url) url that is [outside of home tab scope](https://pr-preview.s3.amazonaws.com/mgiuca/manifest-incubations/pull/95.html#dfn-outside-of-home-tab-scope), the following steps are run:
>
> 1. Behave as `target=_blank`, with _noopener_ and _noreferrer_ specified, creating a new top-level traversable _t_.
t is created in the current window and has the manifest applied.
> 2. Focus _t_.
>
> When [choosing a navigable for a navigation](https://html.spec.whatwg.org/multipage/document-sequences.html#the-rules-for-choosing-a-navigable), if the [URL](https://url.spec.whatwg.org/#concept-url) url is [within home tab scope](https://pr-preview.s3.amazonaws.com/mgiuca/manifest-incubations/pull/95.html#dfn-within-home-tab-scope), the following steps are run:
>
> 1. Let _t_ be the home tab context’s traversable associated with the current window.
> 2. Override the [algorithm](https://html.spec.whatwg.org/multipage/document-sequences.html#the-rules-for-choosing-a-navigable) to return _t_.
> 3. Focus _t_.

mgiuca: @dmurph 

A couple of issues arose trying to implement this:

1. It turns out that "[choosing a navigable](https://html.spec.whatwg.org/multipage/document-sequences.html#the-rules-for-choosing-a-navigable)" is not exported, so for whatever reason, we are not supposed to reference it from outside of HTML. I'm not sure if that means we just shouldn't, or should try and export it, but to avoid shaving that yak first, I think I will just hand-wave it instead of directly referencing this algorithm.
2. We don't actually want to always override the algorithm, only when target=_self. (So for instance, if you click a link with target=<name> then it should still open in the named target.) However, conversely, we want to capture navigations that target this tab, even if they didn't come from this tab (for example, if a child frame does a navigation with target=_parent, into a home URL, we want to capture that into the home tab). So actually I think hooking into the "choosing a navigation" is not the right place. We should let that algorithm run its course, and then _after_ a navigable has been chosen, if it is one of our tabs, override the navigable that actually gets navigated.

I think on account of both of these, it makes more sense to go back to hooking into the algorithm at the "navigate" step. I'll rework it and send it for review.

mgiuca: I've rewritten it to capture the spirit of what we talked about this morning.

I am using "choose a navigable" to get the target=_blank logic, with a red box noting that I can't actually access it (#98).

But primarily, this is now about intercepting "navigation" and doing an "instead of navigating X, navigate Y" type of logic. I think that is more precisely what we want, because then we will intercept if _anything_ in the system tries to navigate one of our tabs.

index.html
mgiuca: Done.

index.html
mgiuca: Done (we're in home_tab/scope_patterns).

index.html
mgiuca: Unclear if this is following any other established pattern. (Most of the "complex" processing steps I can see do not take a manifest member, they just return one as a result. We could do that here, but this way is more flexible and follows the top-level better, so I will choose to keep it unless you object.

It is quite normal to have variable names (in spec) with spaces in them, not underscores.

mgiuca:
@dmurph @loubrett Finished responding to both of your reviews. Thanks for the insightful comments & meeting this morning.

See my comments above - I was not able to do literally what we discussed but I think I have captured the behaviour you wanted. PTAL.

index.html
dmurph: LGTM

dmurph (approved):
LGTM, although nit: the display_mode link in the rendered text seems broken.

loubrett (approved):
LGTM

mgiuca:
> LGTM, although nit: the display_mode link in the rendered text seems broken.

Thanks Dan.

Yes, I mentioned this to you in a chat: we need to wait until w3c/csswg-drafts#7306 gets pushed out to the editor's draft, and then it will automatically resolve itself (because the definition of "display mode" was moved from CSS mediaqueries-5 to Manifest; Manifest updated right away, CSS did not). I don't know when that will be; it was last updated in May 2022, coming up to two years...

Edit: [filed an issue about this](https://github.com/WICG/manifest-incubations/issues/99).

PR #100, opened 2024-05-24
Significant rewrites and updates to the Tabbed Mode explainer.

mgiuca:
This reflects slight design changes that have already made their way into the spec (in particular, the parsing and defaults of scope_patterns: now resolved against the manifest URL instead of scope, and allows any member, not just pathname). Also just clarified a few things and made the writing clearer.

Adds some notes about page navigations being captured, the reason we only allow it in display_override, and discussion around mixing this with other display modes.


loubrett (approved):
Looks good, thanks!

PR #101, opened 2024-06-04
Update scope_extensions explainer to use registrable domains

LuHuangMSFT:
Update scope_extensions explainer to use registrable domains

### DIFF
--- a/scope_extensions-explainer.md
+++ b/scope_extensions-explainer.md
@@ -70,7 +70,7 @@ The `value` field must contain a valid URL.
 
 This format allows for both backward and forward compatibility. For example, a
 new type of entry could filter the paths of an origin added to the extended
-scope. Entries with types are that not recognized by a user agent can be
+scope. Entries with types are that not recognized by a user agent should be
 ignored.
 
 ### Association file
--- a/scope_extensions-explainer.md
+++ b/scope_extensions-explainer.md
@@ -100,7 +100,7 @@ A URL is in the extended scope of a web app if both:
   - That entry has been validated by fetching the
     `<origin>/.well-known/web-app-origin-association` association file with an
     entry matching the app's
-    [identity](https://github.com/philloooo/pwa-unique-id/blob/main/explainer.md).
+    [identity](https://w3c.github.io/manifest/#id-member).
 
 ## Security Considerations
 
--- a/scope_extensions-explainer.md
+++ b/scope_extensions-explainer.md
@@ -66,7 +66,7 @@ The `value` field must contain a valid URL.
 | type   | Behavior |
 |--------|----------|
 | `origin` | The URL in `value` is converted to an [origin](https://html.spec.whatwg.org/multipage/browsers.html#concept-origin-tuple). The `/` path of that origin is added to the extended scope.|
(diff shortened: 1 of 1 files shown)


diekus (approved):

scope_extensions-explainer.md
```diff
@@ -23,40 +23,60 @@ origins.
...
+origins to include in the extended app scope. 
 
-   Example manifest located at `https://example.com/manifest.webmanifest`:
+Example manifest located at `https://example.com/manifest.webmanifest`:
    ```json
    {
      "id": "/",
      "name": "Example",
      "display": "standalone",
      "start_url": "/index.html",
```
dmurph: the start_url cannot be outside the scope, otherwise the manifest doesn't parse the scope member.

```suggestion
     "start_url": "/app/index.html",
```

scope_extensions-explainer.md
```diff
@@ -23,40 +23,60 @@ origins.
...
+     "scope": "/app",
      "scope_extensions": [
-       {"origin": "*.example.com"},
-       {"origin": "example.co.uk"},
-       {"origin": "*.example.co.uk"},
-       {"origin": "*.example.co.uk"}
+       {"type": "registrable_domain", "value": "https://example.com"},
+       {"type": "origin", "value": "https://example.co.uk"},
```
dmurph: This might be weird to do both origin & registrable_domain of .co.uk, as the origin field is redundant then.

Perhaps you could do something like:
```
{ "type": "origin", "value: "https://helpcenter.example-help-center.com" }
```

(and then update the text below as well)

scope_extensions-explainer.md
```diff
@@ -70,13 +90,16 @@ associated origins.
...
-      `scope_extensions` list.
-    - Has an origin with a valid
-      `<origin>/.well-known/web-app-origin-association` association file
-      with an association entry matching the web app's
-      [identity](manifest-identity).
+The `web_app_identity` field must contain a valid [web application
+id](https://w3c.github.io/manifest/#id-member).
+
+### Resulting extended scope
+Let the extended scope of a web app be the set of URLs that:
```
dmurph: rephrase maybe as:

A url is in the extended scope of a web app if both:

* It matches an entry in the `scope_extensions` list.
* That entry has been validated by fetching the `<origin>/.well-known/web-app-origin-association` association file with an entry matching the app's [identity](manifest-identity).

scope_extensions-explainer.md
dmurph: otherwise I have the question - does the registrable domain imply the origin? Or does it not imply the origin? So removing it removes that question. And this can just be an origin example.

dmurph (approved):
Approve to not block, assuming comments above are addressed.

scope_extensions-explainer.md
LuHuangMSFT: Done.

scope_extensions-explainer.md
LuHuangMSFT: After discussion, we decided to have the registrable domain imply inclusion of the origin.  Removed .co.uk origin and added help-center origin example. Updated text below.

scope_extensions-explainer.md
LuHuangMSFT: Done.

PR #103, opened 2024-06-08
Archive translations explainer

christianliebel:
This is now picked up in https://github.com/w3c/manifest/pull/1101.

### DIFF
--- a/README.md
+++ b/README.md
@@ -6,7 +6,6 @@ Specification link: https://wicg.github.io/manifest-incubations/index.html
 # Explainers
 
  - [Scope Extensions for Web Apps](scope_extensions-explainer.md)
- - [Web App Translations](translations-explainer.md)
 
 # Background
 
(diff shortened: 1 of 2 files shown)


dmurph:
Instead of a full delete, can we:
- Create an 'archived' folder to move this into, and
- Add text at the top of the doc to say this is obsolete now and point to your pull request?

christianliebel:
@dmurph Sure, should I move all the other explainers no longer referenced in the README (borderless, note-taking, …) to the _archived_ folder as well?

dmurph:
> @dmurph Sure, should I move all the other explainers no longer referenced in the README (borderless, note-taking, …) to the _archived_ folder as well?

We should only move things that are fully obsolete - I think borderless is being used / referenced still. So perhaps just translations for now?

christianliebel:
SGTM. In that case, this PR should be ready now. 😇

dmurph (approved):
LGTM

PR #107, opened 2024-09-12
Scope extensions explainer update to include path filtering and response header designs

LuHuangMSFT:
Scope extensions explainer update to include path filtering and response header designs


LuHuangMSFT:
Abandoning this PR as I work on more changes in the `scope-extensions-update` branch. Will reopen later.

PR #110, opened 2025-03-27
Create explainer for predictable app updating

Dp-Goog:
Create explainer for predictable app updating, based on discussions in [TPAC minutes](https://www.w3.org/2022/09/13-webapps-minutes.html) and [discussions](https://docs.google.com/document/d/1QDqllh8inOcIkTrblERm4HRKYh8Ce9Lu5S7WhOygs_0/edit?tab=t.0#heading=h.sg3uxnf61udd)


Dp-Goog:
@dmurph PTAL for preliminary checks. Seems like there are some failing IPR checks that I'm not quite sure how to fix. Thanks!

predictable-app-updating.md
```diff
@@ -0,0 +1,182 @@
...
+
+The dialog shows up whenever Chrome sees the new manifest & detects changes. Developers have to accept that every change to SS members could trigger this. They cannot, for example, make a number of incremental changes, and then trigger one update dialog at the end once they are all done.
+
+# **Goals**
+
+The current manifest update process gets the job done, but it could be better in a way so that the problems above can be fixed:
+
+* Provide a consistent way to detect when a manifest update should happen.  
+* Users should not see an update dialog more than necessary to confirm security-sensitive changes.  
+  * The update algorithm should be compatible with tiny image diffs.
```
dmurph: ->   * The update algorithm should work better with tiny image diffs problem above.

Probably good to also say the specifics - where the url doesn't change but the image data does

predictable-app-updating.md
```diff
@@ -0,0 +1,182 @@
...
+
+* If `seen_manifest.update_token` and is changed from the `saved_manifest.update_token,`then a manifest update is triggered.  
+* If an `seen_manifest.update_token` is provided but is unchanged from the app’s `saved_manifest.update_token`, no update is triggered.  
+  * This includes fields outside of `name, short_name` and `icons`, AKA fields that are not tied to the identity of the manifest and is thus not secure.
+
+## Default (non-token-based) update detection
+
+Without the presence of an `update_token` in the manifest, **ONLY** icon updates will be allowed if there are changes in the icon url specified in the manifest.
+
+This will allow icon updates to be enabled on desktop platforms without the existence of the current throttle like behavior.
```
dmurph: The manifest spec doesn't talk about a throttle. Do you mean the throttle that Chromium was required to implement?

predictable-app-updating.md
```diff
@@ -0,0 +1,182 @@
...
+### Pre-requisites:
+
+- An app corresponding to the manifest has already been installed.
+
+# **How does this solve the problem?**
+
+Looking at the goals above and tied it to the proposal:
+
+```
+Provide a consistent way to detect when a manifest update should happen.
```
dmurph: this looks weird (I've never seen this in explainers, using code to list primary reasons? see [w3c](https://tag.w3.org/explainers/) guidance, and the list of examples there)

But if you want it this way then I guess you can keep it.

Dp-Goog: I changed it to use "quotes" instead, what do you think?

predictable-app-updating.md
Dp-Goog: I removed the throttle talk here because that felt more like an implementation detail of Chromium than something that should be on the explainer. Wdyt?

predictable-app-updating.md

dmurph (approved):
LGTM

PR #113, opened 2025-04-06
Add `scope_extensions` member to manifest incubations spec

LuHuangMSFT:
Add `scope_extensions` member to manifest incubations spec

Spec implementation of the `scope_extension` [explainer](https://github.com/WICG/manifest-incubations/blob/gh-pages/scope_extensions-explainer.md).

***
<a href="https://pr-preview.s3.amazonaws.com/WICG/manifest-incubations/pull/113.html" title="Last updated on May 29, 2025, 1:03 AM UTC (d7e2ede)">Preview</a> | <a href="https://pr-preview.s3.amazonaws.com/WICG/manifest-incubations/113/2c02a88...d7e2ede.html" title="Last updated on May 29, 2025, 1:03 AM UTC (d7e2ede)">Diff</a>

### DIFF
--- a/index.html
+++ b/index.html
@@ -1653,7 +1653,7 @@
           <ol>
             <li>
             If |entry| is not an [=ordered map=], or |entry|["scope"] does not
-              exist or is not a [=URL=], continue.</li>
+              [=map/exist=] or is not a [=URL=], continue.</li>
             <li>Let [=URL=] |resolved_scope:URL| be |entry:ordered
               map|["scope"]. </li>
             <li>If |target| is [=URL/within scope=] of |resolved_scope|, return
@@ -1731,12 +1731,12 @@
         </li>
         <li>Set |manifest|["scope_extensions"] to |processedScopeExtensions|.
         </li>
-        <li>If |json|["scope_extensions"] doesn't [=map/exist=] or is not a 
+        <li>If |json|["scope_extensions"] does not [=map/exist=] or is not a 
           [=list=], return.
         </li>
         <li>[=list/For each=] |entry:ordered map| of |json|["scope_extensions"]:
           <ol>
-            <li>If |entry|["type"] or |entry|["origin"] is undefined,
+            <li>If |entry|["type"] or |entry|["origin"] do not [=map/exist=],
               [=iteration/continue=].
             </li>
             <li>If |entry|["type"] is not "origin", [=iteration/continue=].
@@ -1793,19 +1793,19 @@
         <li>Let [=URL=] |origin_url:URL| be the [=URL=] of the [=URL/origin=]
           |extension|["origin"].</li>
         <li>Let [=URL=] |download_url:URL| be the result of [=URL
-          Parser|parsing=] the [=URL=] "/.well-known/web-app-origin-association"
-          using |origin_url| as the base [=URL=].
+          Parser|parsing=] "/.well-known/web-app-origin-association" using 
+          |origin_url| as the base [=URL=].
         </li>
         <li>Let [=ordered map=] |json:ordered map| be the result of a [=fetch=] 
-          using |download_url|. 
+          given |download_url|. 
         </li>
         <li>If the [=fetch=] failed, return an error.
         </li>
-        <li>If |json|[|manifest_id|] does not exist or is not an 
+        <li>If |json|[|manifest_id|] does not [=map/exist=] or is not an 
           [=ordered map=], return an error.</li>
         <li>Let [=string=] |scope:string| be "/".</li>
-        <li>If |json|[|manifest_id|]["scope"] exists and is a [=string=], set
-          |scope| to |json|[|manifest_id|]["scope"]. 
+        <li>If |json|[|manifest_id|]["scope"] [=map/exists=] and is a
+          [=string=], set |scope| to |json|[|manifest_id|]["scope"]. 
         </li>
         <li>Let [=URL=] |resolved_scope:URL| be the result of [=URL
           Parser|parsing=] |scope| with |extension|["origin"] as the base 
--- a/index.html
+++ b/index.html
@@ -1809,6 +1809,18 @@
       <p>The user agent MAY choose to apply <dfn>apply extended scope</dfn> in
         some scenarios and [=manifest/scope=] in others. 
       </p>
+      <p>
+        If the [=application context=]'s [=navigable/active document=]'s [=URL=]
+        is not [=manifest/within scope=] but is [=within extended scope=], the
+        user agent SHOULD provide UI that allows the user to determine the
+        [=URL=] or at least its [=origin=], including whether it is served over
+        a secure connection. This UI SHOULD differ from any UI used when the
+        [=URL=] is not [=manifest/within scope=] of the [=application
+        context=]'s [=Document/processed manifest=], in order to make it obvious
+        that the user is still navigating the intended contents of the
+        application but also keep the user informed of the privacy and security
+        implications of navigating to a different [=origin=].
+      </p>
     </section>
     </section>
     <section>
--- a/index.html
+++ b/index.html
@@ -1657,8 +1657,8 @@
               map|["scope"]. </li>
             <li>If |target| is [=URL/within scope=] of |resolved_scope|, return
               `true`.</li>
-            <li>Return `false`.</li>
           </ol>
+          <li>Return `false`.</li>
       </ol>
       <aside class="note">
         The [=manifest's=] `scope` member only allows URLs to be considered for scope
--- a/index.html
+++ b/index.html
@@ -1626,8 +1626,8 @@
       <p>
         Each <dfn>scope extension</dfn> extends the [=manifest/navigation scope
         of a manifest=] by describing [=URLs=] that should be treated as being 
-        [=manifest/within scope=]. [=Scope extensions=] can only add URLs to 
-        scope, not remove them.
+        [=manifest/within scope=]. [=Scope extensions=] can add URLs to scope 
+        but not remove them.
       </p>
       <p>
         The [=manifest's=] <code><dfn data-export="" data-dfn-for="manifest">
@@ -1693,6 +1693,26 @@
 
       <section>
       <h2>Processing the scope_extensions member</h2>
+      
+      <p>
+        An <dfn>extension type</dfn> represents a method of selecting URLs to be
+        added to the scope as an extension. 
+      </p>
+      <p>
+        This specification defines the following [=extension types=]:
+      </p>
+      <dl>
+        <dt>
+          <dfn class="export" data-dfn-for="extension type">origin</dfn>
+        </dt>
+        <dd>
+          Adds URLs from a single [=origin=] as a [=scope extension=].
+        </dd>
+      </dl>
+      <p>
+        The <dfn>extension types list</dfn> is the [=list=] 
+        « "[=extension type/origin=]" ».
+      </p>
       <p>
         To <dfn>process the `scope_extensions` member</dfn>, given
         [=ordered map=] |json:ordered map| and [=ordered map=]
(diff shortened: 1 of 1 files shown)


LuHuangMSFT:
@diekus @HowardWolosky PTAL.

LuHuangMSFT:
Working on an update to this PR to define "extended scope" as a new concept as a union of "scope" as defined in the original W3C spec AND new scope extensions. This way we can avoid overriding definitions in the original spec (monkey-patching problem).

index.html
```diff
@@ -1615,6 +1618,179 @@ <h3>
...
     </section>
+    <section>
+      <h3>
+        `scope_extensions` member
+      </h3>
+      <p>
+        Each <dfn>scope extension</dfn> extends the [=manifest/navigation scope
+        of a manifest=] by describing [=URLs=] that should be treated as being 
+        [=manifest/within scope=]. [=Scope extensions=] can add URLs to scope 
+        but not remove them.
```
dmurph: We can probably remove the last sentence and add 'additional'?
```suggestion
        Each <dfn>scope extension</dfn> extends the [=manifest/navigation scope
        of a manifest=] by describing additional [=URLs=] that should be treated as
        being [=manifest/within scope=].
```

index.html
```diff
@@ -1615,6 +1618,179 @@ <h3>
...
+      </p>
+      <p>
+        The [=manifest's=] <code><dfn data-export="" data-dfn-for="manifest">
+          scope_extensions</dfn></code> member represents a list of an
+          application's desired [=scope extensions=].
+      </p>
+      <p>
+        The user agent MUST [=process the scope_extensions member=] and 
+        [=validate scope extensions=] before applying them.
```
dmurph:
```suggestion
        [=validate scope extensions=] before they can be [=apply scope extensions/applied=] to allow more [=URLs=] to be [=manifest/within scope=].
```

index.html
```diff
@@ -1615,6 +1618,179 @@ <h3>
...
+      <p>
+        A [=web-app-origin-association=] file is expected to be downloadable
+        from <code>[origin]/.well-known/web-app-origin-association</code>.
+      </p>
+    </section>
+    <section>
+      <h2>Applying scope extensions</h2>
+      <p>The user agent MAY apply extended scope to some features but 
+        not others.
+      </p>
```
dmurph: We should `dfn` the application, and define the algorithm? Ideally we can reference the `[=manifest/within scope=]` algorithm but give it different 'input arguments', as we have the same 'scope' member here (and that's what we want in the future - same algorithm, but different inputs.

index.html
```diff
@@ -1615,6 +1618,179 @@ <h3>
...
+            <li>If <var>entry</var>["type"] or <var>entry</var>["value"]
+              is undefined, [=iteration/continue=].
+            </li>
+            <li>If |entry|["type"] is not `origin`, 
+              [=iteration/continue=].
+            </li>
+            <li>If <var>entry</var>["value"] is not a valid [=URL=], 
+              [=iteration/continue=].
+            </li>
+            <li>Set <var>entry</var>["value"] to its own [=origin=].
```
dmurph: delete this step?

index.html
```diff
@@ -1615,6 +1618,179 @@ <h3>
...
+      <section>
+      <h2>Validating scope extensions</h2>
+      <p>
+        To <dfn>validate scope extension</dfn>, given [=ordered map=] 
+        |extension:ordered map|, [=URL=] |manifest_id:URL| and [=URL=] 
+        |resolved_scope:URL|:
+      </p>
+      <ol class="algorithm">
+        <li>Set |resolved_scope| to empty URL.</li>
+        <li>Let |download_url| be the result of constructing
```
dmurph: Probably we want to resolve the url, using manifest_url as the 'base', put that into a variable

index.html
```diff
@@ -1615,6 +1618,179 @@ <h3>
...
+              [=iteration/continue=].
+            </li>
+            <li>If <var>entry</var>["value"] is not a valid [=URL=], 
+              [=iteration/continue=].
+            </li>
+            <li>Set <var>entry</var>["value"] to its own [=origin=].
+            </li>
+            <li>Validate |entry| using the steps from [=validate scope 
+              extensions=], given [=ordered map=] |entry:ordered map|, and 
+              [=URL=] |resolved_scope:URL|.
```
dmurph: we also need to call the algorithm with our manifest_id?

dmurph (commented):
Initial review

index.html
LuHuangMSFT: I added a new dfn at the top of the whole section to specify that extension scope is original manifest scope + URLs in extensions. Does that accomplish the same thing? 

I'm using "Application" section for things like UA should keep users informed of the security origin, etc.

index.html
LuHuangMSFT: In the explainer we required that `origin` value be a URL, but that we'll truncate to its origin. In implementation, we just try to construct a url::Origin directly from the string.

index.html
LuHuangMSFT: Are you referring to the construction of the download_url or resolved_scope here?

index.html
LuHuangMSFT: Added manifest_id to params.

index.html
```diff
@@ -1615,6 +1618,211 @@ <h3>
...
+        scope=] of the manifest or [=matches a validated scope extension=]. 
+      </p>
+      <p>
+        A [=URL=] |target:URL| <dfn>matches a validated scope extension</dfn> if
+        the following algorithm returns `true`, given |target| and [=ordered map=]
+        |validated_scope_extension:ordered map|:
+      </p>
+      <ol class="algorithm">
+        <li>Let [=URL=] |resolved_scope:URL| be
+          |validated_scope_extensions:ordered map|["scope"]. </li>
```
dmurph:
```suggestion
          |validated_scope_extension:ordered map|["scope"]. </li>
```

index.html
```diff
@@ -1615,6 +1618,211 @@ <h3>
...
+        <li>If |target| is [=URL/within scope=] of |resolved_scope|, return
+          `true`.</li>
+        <li>Return `false`.</li>
+      </ol>
+      <aside class="note">
+        The [=manifest's=] `scope` member only allows URLs to be added to scope
+        from a single [=origin=]. The `scope_extensions` member allows URLs to
+        be added from multiple [=origins=].
```
dmurph: 'added' might  be a bit weird here

```suggestion
        The [=manifest's=] `scope` member only allows URLs to be considered for scope
        from a single [=origin=]. The `scope_extensions` member allows URLs to
        be considered from multiple [=origins=].
```

index.html
```diff
@@ -1615,6 +1618,211 @@ <h3>
...
+        [=validate scope extensions=] before it can [=apply extended scope=] 
+        to allow additional [=URLs=] to be [=within extended scope=].
+      </p>
+      <p>
+        A [=URL=] |target:URL| is <dfn>within extended scope</dfn> of a
+        |manifest:processed manifest| if the |target| is [=manifest/within
+        scope=] of the manifest or [=matches a validated scope extension=]. 
+      </p>
+      <p>
+        A [=URL=] |target:URL| <dfn>matches a validated scope extension</dfn> if
```
dmurph, line 1646: We never really talk about iterating the list anywhere....

Should we do that in this algorithm? Instead of just one, we iterate them, check each one?

index.html
```diff
@@ -1615,6 +1618,211 @@ <h3>
...
+            <li>If |entry|["type"] or |entry|["origin"] is undefined,
+              [=iteration/continue=].
+            </li>
+            <li>If |entry|["type"] is not "origin", [=iteration/continue=].
+            </li>
+            <li>If |entry|["origin"] is not a [=string=],
+              [=iteration/continue=].
+            </li>
+            <li>Set |entry|["origin"] to the result of parsing it as an
+              [=origin=].
```
dmurph: There is a spec algorithm for this we should reference. 

The origin of a url is https://url.spec.whatwg.org/#origin

parsing is https://url.spec.whatwg.org/#concept-url-parser


```suggestion
            <li>Set |origin url| to the result of [=URL Parser|parsing=] |entry|["origin"], using the |manifest url| as the base URL]
            </li>
            <li>Set |entry|["origin"] to the [=URL/origin=] of the |origin url|.
```

You'll have to add the manifest url as an argument too. I'm not totally sure offhand about the origin link, but yeah - this should be linked to url parsing & url spec definitions.

index.html
```diff
@@ -1615,6 +1618,211 @@ <h3>
...
+            </li>
+            <li>Let [=URL=] |resolved_scope:URL| be an empty [=URL=]. </li>
+            <li>Validate |entry| using the algorithm from [=validate scope
+              extensions=], given [=ordered map=] |entry:ordered map|, [=URL=]
+              |manifest_id:URL|, and [=URL=] |resolved_scope:URL|.
```
dmurph: why do we pass it the scope? Can it return that?

Looking at the manifest spec, we can say that things 'return an error', and we can check for that instead of 'validation'.

index.html
```diff
@@ -1615,6 +1618,211 @@ <h3>
...
+      </section>
+      <section>
+      <h2>Validating scope extensions</h2>
+      <p>
+        To <dfn>validate scope extensions</dfn>, given [=ordered map=] 
+        |extension:ordered map|, [=URL=] |manifest_id:URL| and [=URL=] 
+        |resolved_scope:URL|:
+      </p>
+      <ol class="algorithm">
+        <li>Set |resolved_scope| to empty URL.</li>
```
dmurph: is it normal to accept an argument like this as an 'out' argument? Do other algorithms in the manifest spec do this?

index.html
```diff
@@ -1615,6 +1618,211 @@ <h3>
...
+        <li>Let [=URL=] |download_url| be the result of parsing
```
dmurph: same as above - you can reference the parsing algorithm like URL parsing in the manifest spec does

index.html
```diff
@@ -1615,6 +1618,211 @@ <h3>
...
+        </li>
+        <li>If the download fails, return false.
+        </li>
+        <li>If |json|[|manifest_id|] does not exist or is not an 
+          [=ordered map=], return false.</li>
+        <li>Let [=string=] |scope_path:string| be "/".</li>
+        <li>If |json|[|manifest_id|]["scope"] exists and is a [=string=], set
+          |scope_path| to |json|[|manifest_id|]["scope"]. 
+        </li>
+        <li>Set |resolved_scope:URL| be the result of parsing |scope_path| with
```
dmurph: same

index.html
```diff
@@ -1615,6 +1618,211 @@ <h3>
...
+    </section>
+    <section>
+      <h2>Applying scope extensions</h2>
+      <p>The user agent MAY choose to apply <dfn>apply extended scope</dfn> in
+        some scenarios and [=manifest/scope=] in others. 
+      </p>
+      <p>
+        If the [=application context=]'s [=navigable/active document=]'s [=URL=]
+        is not [=manifest/within scope=] but is [=within extended scope=], the
+        user agent SHOULD provide UI that allows the user to determine the
```
dmurph, line 1825: this is a contractual thing. I don't think Chromium implementation would be in-spec. What do you mean by the 'determine the url or at least it's origin" here.

Where did this requirement come from?

dmurph (changes requested):

index.html
dmurph: um.... `download_url` probably, and also probably `resolved_scope` too? Basically when we make a url, if it's from a string, we have to use a spec algorithm to do it.

index.html
LuHuangMSFT: [[=URL/within scope=]](https://www.w3.org/TR/appmanifest/#dfn-within-scope) is the algorithm to define [[=manifest/within scope]](https://w3c.github.io/manifest/#dfn-within-scope-0). We now use [=URL/within scope=] to define [=within extended scope=] as well (indirectly, through [=matches a validated scope extension=].

index.html
LuHuangMSFT: Wait, these origin input strings aren't supposed to be relative URLs. I'll reference the parsing algo but we don't need a base URL here.

index.html
LuHuangMSFT: Changed to return error or resolved scope URL.

index.html
LuHuangMSFT: Addressed by not doing this anymore.

index.html
LuHuangMSFT: Done.

index.html
LuHuangMSFT: Done.

index.html
LuHuangMSFT: Done for both.

index.html
LuHuangMSFT: Should be addressed by this step:
```
            <li>Let [=URL=] |origin_url:URL| be the result of [=URL
              Parser|parsing=] |entry|["origin"].
```

LuHuangMSFT:
PTA(another)L. Comments addressed. Thanks!

index.html
dmurph: SGTM

dmurph (approved):

PR #115, opened 2025-04-28
Clearly explain update_token behavior differences in explainer for predictable app updating.

Dp-Goog:
This PR updates the predictable app updating explainer in a few ways:
- Updates the old table in the explainer about the behavior differences in between the saved and seen `update_token` field in the manifest to provide more clarity on the proposed behavior.
- Ensures that the impact of the main proposal is clearer to the developer and the user, based on [TAG feedback](https://github.com/w3ctag/design-reviews/issues/1076).


predictable-app-updating.md
```diff
@@ -4,34 +4,41 @@ Author: Dibyajyoti Pal (dibyapal@chromium.org)
 
 # **Introduction**
 
-This explainer proposes a way to have PWAs fully update their identities in a safe and resourceful manner, to further bridge the gap between PWAs and native apps. This is done by updating the [manifest spec](https://www.w3.org/TR/appmanifest/#web-application-manifest) to have a specific field for making updates more deterministic by the developer, and for providing a consistent experience. The proposal attempts to do so in a way that:
+This explainer proposes a way to have PWAs fully update their identities in a safe and resourceful manner, to further bridge the gap between PWAs and native apps. This is done by updating the [manifest spec](https://www.w3.org/TR/appmanifest/#web-application-manifest) to have a specific field for making updates more deterministic by the developer, and for providing a consistent experience across different user agents. The proposal attempts to do so in a way that:
```
dmurph: This isn't quite correct & maybe a reason for some confusion. This doesn't propose a way for identity updating, this proposes a more consistent for developers and backwards compatible way for updates to be checked.

predictable-app-updating.md
```diff
@@ -4,34 +4,41 @@ Author: Dibyajyoti Pal (dibyapal@chromium.org)
...
 
 1. Uses less resources, making network usage more efficient.
```
dmurph: 1. Specifies the update check algorithm to ensure the user agent can use minimal network resources

predictable-app-updating.md
```diff
@@ -4,34 +4,41 @@ Author: Dibyajyoti Pal (dibyapal@chromium.org)
 
 # **Introduction**
 
-This explainer proposes a way to have PWAs fully update their identities in a safe and resourceful manner, to further bridge the gap between PWAs and native apps. This is done by updating the [manifest spec](https://www.w3.org/TR/appmanifest/#web-application-manifest) to have a specific field for making updates more deterministic by the developer, and for providing a consistent experience. The proposal attempts to do so in a way that:
+This explainer proposes a way to have PWAs fully update their identities in a safe and resourceful manner, to further bridge the gap between PWAs and native apps. This is done by updating the [manifest spec](https://www.w3.org/TR/appmanifest/#web-application-manifest) to have a specific field for making updates more deterministic by the developer, and for providing a consistent experience across different user agents. The proposal attempts to do so in a way that:
 
 1. Uses less resources, making network usage more efficient.  
 2. Prevent user confusion by showing the update UX less often.
+3. Allows user agents who haven't yet implemented manifest updating to do so in a different way.
```
dmurph: Delete

predictable-app-updating.md
```diff
@@ -4,34 +4,41 @@ Author: Dibyajyoti Pal (dibyapal@chromium.org)
...
+
+One thing to note here is that installed PWAs are inherently different from how the rest of the Web Platform works. While the origin is the security boundary of the Web Platform, the installed experience has [different security parameters](https://www.w3.org/TR/appmanifest/#dfn-security-sensitive-members).
```
dmurph: Why are you saying this? What was the root of that message they sent? They were saying that it's weird to need a token for something like this on the web, why not just make it so icon updates don't happen unless the URL changes?

The message here is that manifest properties that are security sensitive<put link> are shown by the user agent instead of an origin, so changes to these properties must be handled carefully and the manifest update algorithm<link> already specifies that UX must be shown to the user when they do.

The point of this explainer (please make sure it's presented as such) is that this is specifying 'when' to consider update, and specifying it in a way that accomplishes the above 2 goals.

predictable-app-updating.md
```diff
@@ -4,34 +4,41 @@ Author: Dibyajyoti Pal (dibyapal@chromium.org)
...
+[PWAs](https://web.dev/explore/progressive-web-apps) are app experiences built on the web, and like native apps, they support updating themselves. Developers have the option of changing any field in the manifest (including the [security sensitive fields](https://www.w3.org/TR/appmanifest/#dfn-security-sensitive-members)), and the user agent can then apply the changes in a way as defined in the [manifest updating spec](https://www.w3.org/TR/appmanifest/#updating). For example, in an effort to prevent [phishing risks](#phishing), some user agents like Chromium shows an update dialog to the user.
```
dmurph: mention that they SHOULD show UX.  This line from the spec is a requirement for user agents!

```
Instead, user agents SHOULD present changes to [security-sensitive members](https://www.w3.org/TR/appmanifest/#dfn-security-sensitive-members) with appropriate management options, so the user can make an informed decision about updating the web application.
```

predictable-app-updating.md
```diff
@@ -125,6 +162,19 @@ The most network heavy traffic is downloading icons, which will only happen when
...
+
+Add the concept of versioning to the web app manifest, so that the developer can change "versions" of their installed experiences to trigger updates. This has been proposed earlier in the manifest spec multiple times (like [manifest#1036](https://github.com/w3c/manifest/issues/1036) and [manifest#1157](https://github.com/w3c/manifest/issues/1157)).
+
+Pros:
+- This would also help support this [FR about changelogs](https://github.com/w3c/manifest-app-info/issues/1).
+
+Cons:
+- As commented in [the issue](https://github.com/w3c/manifest/issues/1036), there is general conflation about what this field will be used for. The term is highly overloaded, and it is [hard to get consensus on the data type](https://github.com/w3c/manifest/issues/446#issuecomment-904359501) and what its purpose is. Compared to that, `update_token` is highly confined to a specific use-case, and is simpler to reason
+about.
+- Leads to complications around supporting [version downgrades](https://github.com/w3c/manifest/issues/446#issuecomment-905725612).
```
dmurph: be more specific. Leads to complications about interpretation of the field, expectations around 'upgrades' and 'downgrades', etc.

predictable-app-updating.md
```diff
@@ -125,6 +162,19 @@ The most network heavy traffic is downloading icons, which will only happen when
...
+- This does not actually empower developers to have more control over app updates.
```
dmurph: delete?

predictable-app-updating.md
```diff
@@ -140,11 +190,9 @@ Cons:
 
 While this solution is simple to implement, it breaks existing update behavior, and is also cumbersome for developers, who would have to update where they serve their manifest from every time a change is made.
 
-## **Future Considerations**
-
-### Allow developers to use javascript to trigger a pending update
+## Use [Cache-Control:Immutable](https://caniuse.com/mdn-http_headers_cache-control_immutable) for static icon urls
 
-This proposal allows developers to control when an update happens in general for users of an old version, but not at a specific time in the user experience of their site. In the future a javascript API could be used similar to the `beforeinstallprompt` API, allowing the developer to put UX on their site to trigger the update dialog.
+While `Cache-Control:Immutable` can be used to help solve the use-case for developers that do not change their icon urls, it does not really empower developers to choose when to trigger updates if they want to do so.
```
dmurph: Ah I think I didn't explain this suggestion well. They were suggesting we only do the non-token algorithm. The alternative is that we change our update detection to only download icons if it's a new url.

BTW, we can communicate this behavior (the non-token update detection algorithm) as mimicing the `Cache-Control:Immutable` behavior. You should do this above, when we describe that algorithm.

The con is that this doesn't allow backwards compatibility with existing PWAs, especially on Android, that can update by only changing the icon bitmap and not the url.

dmurph (changes requested):

predictable-app-updating.md
Dp-Goog: Done, wdyt?

predictable-app-updating.md

predictable-app-updating.md

predictable-app-updating.md
Dp-Goog: Updated.

predictable-app-updating.md
Dp-Goog: Updated some text in the background to be:

```
For example, changes to the security sensitive manifest properties **require** the user agent to explicitly get user permission, which is why Chromium shows an update dialog to the user. This helps prevent [phishing risks](#phishing).
```

predictable-app-updating.md

predictable-app-updating.md

predictable-app-updating.md

dmurph (approved):

PR #117, opened 2025-05-07
Update predictable app updating explainer, remove update_token

Dp-Goog:
This PR does the rewords the predictable app updating explainer by proposing an alternative icon_url based solution, moving the previous `update_token` solution as an alternative as per [TAG feedback](https://github.com/w3ctag/design-reviews/issues/1076).

Also changes the answer to a single question in the self-review questionnaire about identifiers in the specification being exposed to the web, which it currently does not.


predictable-app-updating.md
```diff
@@ -57,58 +57,25 @@ Chrome on Android solved it with a stop gap where PWA updates were automatic if
...
 * [**Preventing unnecessary user interruption**](#useint): Users should not see an update dialog more than necessary to confirm security-sensitive changes.
 * [**Consistency**](#consistency): Provide a consistent way to detect when a manifest update should happen, which can be implemented across all user agents.
 * [**User agent flexibility**](#uaf): It should be possible for user agents to use their judgement to block updates for known bad sites, allow known trusted apps to update without UX, or allow tiny visual changes to icons without requiring UX.
 * [**Reduce network traffic**](#traffic): Unnecessary network traffic should be minimized.
-* [**Prevent manifest id foot-gun**](#footgun): Encourage developers to set the manifest 'id' field, preventing a known [foot-gun](https://github.com/w3c/manifest/issues/1148).
 
-# [**Proposal: Introduce 'update\_token', ignore icon changes by default**](#proposal)
+# [**Proposal: Perform app updates silently, icon updates only when icon url changes**](#proposal)
```
dmurph:
```suggestion
# [**Proposal: Perform non-security-sensitive field updates silently, icon updates only when icon url changes, and security-sensitive field updates are optional (or user-agent mediated)**](#proposal)

```

predictable-app-updating.md
```diff
@@ -57,58 +57,25 @@ Chrome on Android solved it with a stop gap where PWA updates were automatic if
...
-  "display": "standalone",
-   ....
-   "update_token": "foo1"
-}
-
-```
+The proposal is to:
+- Allow the updating of non-security sensitive members silently.
+- Allow updating security sensitive members like name with an UX shown by the user agent.
+- For icons, the update dialog shows up only if the [icon url in the manifest](https://www.w3.org/TR/appmanifest/#icons-member) has changed, similar to [Cache-Control:Immutable](https://caniuse.com/mdn-http_headers_cache-control_immutable) behavior.
```
dmurph:
```suggestion
- For icons, only consider an icon updated for a given size if the [icon url in the manifest](https://www.w3.org/TR/appmanifest/#icons-member) has changed, similar to [Cache-Control:Immutable](https://caniuse.com/mdn-http_headers_cache-control_immutable) behavior.
```

predictable-app-updating.md
```diff
@@ -143,25 +110,32 @@ updateButton.addEventListener('click', async () => {
 });
 ```
 
+This would be the ultimate solution, because it gives the developer "full" control over the manifest update process.
```
dmurph:
```suggestion
This would be an ultimate solution for developer control, because it gives the developer "full" control over the manifest update process. However, allowing the security sensitive members to be mediated by the user agent allows the update process to be less blocking, so the main developer complaint of not wanting blocking dialogs for users is mitigated.
```

predictable-app-updating.md
```diff
@@ -143,25 +110,32 @@ updateButton.addEventListener('click', async () => {
...
+
 # **How does this solve the problem?**
 
 Let’s review the goals again to see how this proposal meets them: 
 
 > Goal: [`Consistency`](#consistency), [`User Agent Flexibility`](#uaf)
 
-The presence of a different value of `update_token` compared to the one saved, and icon urls changing are the only 2 use-cases that can trigger a manifest update. Since this will be specified in the manifest, user agents can choose to implement their own behavior around this.
+Since the change in an icon url is the **`signal`** that a manifest update can be triggered, this allows user agents to implement their own behavior around it. This will be [speced in the manifest](https://www.w3.org/TR/appmanifest/#updating) for consistency across multiple
+user agents.
```
dmurph: It might be nice to add:
```md
+ The algorithm will treat a url as 'updated' or 'changed' by looking at the previously seen manifest's specified icon for each size and purpose.
```

And then later maybe something like:

```
Note: This means that a developer who specifies the same icon url, but for a different size, is technically eligible for a re-download by the user agent, as the user agent may cache icons by downloading them at the sizes specified by the developer. This behavior keeps the implementation & expectations simple.
```

predictable-app-updating.md
```diff
@@ -172,6 +146,14 @@ Cons:
 - The term is highly overloaded, and it is [hard to get consensus on the data type](https://github.com/w3c/manifest/issues/446#issuecomment-904359501) and what its purpose is. Compared to that, `update_token` is highly confined to a specific use-case, and is simpler to reason about.
 - Leads to complications around supporting a lot of things like [version downgrades](https://github.com/w3c/manifest/issues/446#issuecomment-905725612), upgrades as well as [ overall interpretation of the field](https://github.com/w3c/manifest/issues/1036).
 
+## Use an `update_token` field, only for manifest updates
+
+This is kind of similar to the above suggestion of using a [version field](#version), with the only difference being that the field is highly confined to be of a single data type like string, and has a single purpose, which is to signal that the manifest needs updating.
+
+Cons:
+- This requires an extra image diff algorithm to determine if icon bits have changed if the url is still the same during the update process.
```
dmurph, line 155: Important issue here that goes against the goals - Potential foot-gun: If the developer doesn't update the token, and gradually changes the app, then they will have different app functionalities in the wild possibly without knowing.

dmurph (changes requested):

predictable-app-updating.md

predictable-app-updating.md

predictable-app-updating.md
Dp-Goog: I like this wording, done.

predictable-app-updating.md
Dp-Goog: I added them as a behavior just under the proposal as that seems like the place to put this in (AKA what we're proposing and how it will work). PTAL!

predictable-app-updating.md
Dp-Goog: Done

dmurph (approved):
LGTM

Dp-Goog:
Thanks for the quick reviews here Dan!

PR #119, opened 2025-09-05
Update unframed explainer with url_patterns proposal

edman:
This renames `borderless` to `unframed` and updates the explainer with `url_patterns`.

The explainer now extends `display_override` to accept JSON object with `display` and `url_patterns` fields.

Apps can then declare which URLs should be unframed via `url_patterns`. This enables mixed scenarios where some windows are unframed and some are normal.

The solution is generic and can be applied in other display modes where it makes sense, like `window-controls-overlay` or `tabbed`.

***
<a href="https://pr-preview.s3.amazonaws.com/edman/borderless-explainer-in-manifest-incubations/pull/119.html" title="Last updated on Dec 5, 2025, 2:33 PM UTC (7fca618)">Preview</a> | <a href="https://pr-preview.s3.amazonaws.com/WICG/manifest-incubations/119/94fc92c...edman:7fca618.html" title="Last updated on Dec 5, 2025, 2:33 PM UTC (7fca618)">Diff</a>


borderless-explainer.md
```diff
@@ -1,476 +1,217 @@
...
+```
 
-Some implementations of PWAs, e.g. Chrome and Edge, show a three-dot menu
-button. This gives users access to capabilities like for example extensions,
-copy URL, open in browser, zoom, print, find, cast, cut, copy, paste etc
-depending on the browser.
+The `borderless_url_patterns` field is a new web app manifest field we propose
+for this feature. It accepts URL patterns as described in the
+[URL pattern spec](https://urlpattern.spec.whatwg.org) and describes the set of
+URLs that should be borderless.
```
dmurph: I'm a little worried about lack of specification for when the display/display_override value is used, vs the borderless value is used.

One idea would be to continue the same concepts, and introduce, say, display mode definitions like so:

```json
{
...
"custom_display_modes": {
  "x-borderless": {
    "type": "borderless"
    # the parsing of these would be based on the type:
    "url_patterns": [
      "/some/path/*",
      { "pathname": "/some/other/path/*" }
    ]
},

...
# and then this can be specified in the display_override list

"display_override": ["x-borderless"],
"display": "standalone",
...
}
```

That way, we can define a thing that is then usable in the display_override fallback list, which is very well defined.

Also, this allows for future customization of display modes, where, say, we allow the developer to modify other display mode configurations (and because of how the display_override fallback list allows for user agents to decide whether those display modes are supported or not, it gives developers control for their requirements)

This might be a little overkill .... @mek I'm interested in your thoughts here too. Perhaps just having the existing proposal and also 'borderless' specified in the display_override is good enough, where the spec says that borderless display mode will use configuration put in the manifest. I just really like the display mode definition system for some reason....

edman: Thanks Dan for reviewing the PR!

Both options you suggested sound good on my side.

Do you know of other scenarios that could benefit from `custom_display_modes`?

And are you assuming that the type of `x-borderless` in your example is somehow fixed? Perhaps a tagged union based on the `type` field?

dmurph: (related idea from an old proposal: https://github.com/WICG/display-override/blob/main/explainer.md#custom-display-mode-names-with-display-modifiers-style-specification)

x-borderless - that's just the developer creating a name for it.
- (in my head) we could say it has to start with "x-" to make it not conflict with future display modes
- it could be "x-first-supported-display-mode" or whatever the dev wants - that just creates the name, and that can be used in display_override

edman:
Hi @dmurph, this PR is ready for review.

dmurph (approved):

PR #121, opened 2025-09-09
Add app migration explainer draft

mkruisselbrink:


mkruisselbrink:
@dmurph can you PTAL?

pwa-migration-explainer.md
```diff
@@ -0,0 +1,199 @@
...
+
+However, there is a critical security consideration for `"force"` migrations. To prevent a developer from forcing a user to accept a potentially deceptive name or icon change, the browser will enforce a separation of concerns. During a **forced migration, the browser will only apply the URL change**. If the new manifest also contains changes to security-sensitive fields like `name` or `icon`, the browser will defer these changes. After the migration is complete, these deferred changes will be processed as a standard, ignorable update, allowing the user to review them separately and preserving user control over the app's identity.
+
+## Alternatives considered
+
+### Require a 'cooling off' period
+
+A mandatory waiting period (e.g., 2 weeks) after a migration is declared was considered to help prevent malicious activity if a site is compromised. However, this was deemed overly complex and potentially not useful, as there's no guarantee a developer would notice a compromise within that period. The **same-site restriction** is a much stronger and more reliable security guarantee, as it ensures both origins are controlled by the same entity.
+
+### Only require a one-way reference
```
dmurph: I think the current design allows a one-way reference from the 'new' app to the old by design.

probably this should be "Requiring a two-way reference" alternative, and the 'con' is for difficult scenarios where the redirection is happening (without specifying an install_url), or just that it's a really old install and the dev doesn't maintain any of the 'old' system / manifest stuff anymore. Also performance / speed of needing to fetch the old app manifest if we only see the new one?

pwa-migration-explainer.md
```diff
@@ -0,0 +1,199 @@
...
+
+We considered only requiring the `migrate_to` field in the old manifest. This would simplify the developer's work slightly. However, it would make it impossible for the browser to verify the migration if the user navigates directly to the new app's origin first. The two-way handshake ensures the migration can be discovered and verified from either direction, providing a more robust solution.
+
+### Require an Explicit `id` in both manifests
```
dmurph, line 179: I'm still.... I actually still want this. not having an 'id' causes a bunch of problems, can we at least require an id in the new manifest?

pwa-migration-explainer.md
```diff
@@ -0,0 +1,199 @@
...
+### Goals
+
+*   **Seamless User Experience:** Users should be able to transition to a new PWA origin with a simple, non-disruptive update flow.
+*   **Developer Enablement:** Provide a clear mechanism for developers to migrate their PWAs for reasons like rebranding (e.g., "ConnectApp" to "ConnectX"), SEO optimization (e.g., `photo-app.example.com` to `photos.example.com/app`), or backend architecture changes (e.g., `mail.example.com` to `app.example.com`).
+*   **Permission Persistence:** User-granted permissions (e.g., notifications, location) should be securely and transparently migrated to the new origin where possible.
+*   **Security:** The migration process must be secure, preventing malicious takeovers or phishing attacks.
+
+### Non-goals
+
+*   **Cross-Site Migration:** This feature will not support migration between origins that are not "same-site" (e.g., from `example.com` to `another-site.com`).
```
tomayac: Is it worth saying why this is a non-goal? The arguably biggest PWA migration (`twitter.com` → `x.com`) would not have profited from this proposal had it existed at the time.

pwa-migration-explainer.md
mkruisselbrink: oops, yeah, that's what I meant to include; changed.

pwa-migration-explainer.md
mkruisselbrink: I'm pretty neutral on requiring that or not, so sure, I updated the explainer to require id for migrate_from and added an alternative to not require it at all.

pwa-migration-explainer.md
mkruisselbrink: Added a bit more context and clarified why not and that we might want to explore this in the future

pwa-migration-explainer.md
```diff
@@ -0,0 +1,210 @@
...
+  "migrate_to": {
+    "id": "https://social.example.com/",
+    "install_url": "https://social.example.com/download?usp=migrate"
+  }
+}
+```
+
+Once the migration is discovered (via either method), the browser will present the user with an update prompt. The developer can control the user experience with the optional `behavior` field in the `migrate_from` object:
+*   `"suggest"`: The user is (passively) notified of the update but can ignore it.
+*   `"force"`: The next time the user launches the app, they are presented with a blocking dialog requiring them to either migrate or uninstall the app.
```
dmurph:
```suggestion
*   `"force"`: The next time the user launches the app (after a manifest is parsed with this behavior specified), they are presented with a blocking dialog requiring them to either migrate or uninstall the app.
```

maybe to be clear about exactly when?

pwa-migration-explainer.md
```diff
@@ -0,0 +1,210 @@
...
+
+### Never require an `id`
+
+We also considered not requiring an `id` field in the manifest at all for migrations. The migration process itself does not technically depend on the `id` field. The primary motivation for requiring it is to encourage the adoption of a best practice that provides a stable identity for a PWA, preventing potential issues in the future. However, making this a requirement for an unrelated feature could be seen as using the migration mechanism to enforce an orthogonal best practice.
```
dmurph:
```suggestion
We also considered not requiring an `id` field in the manifest at all for migrations. The migration process itself does not technically depend on the `id` field. The primary motivation for requiring it is to encourage the adoption of a best practice that provides a stable identity for a PWA, preventing [common unrecoverable issues in the future](https://docs.google.com/presentation/d/1QqriRwxqzo8Dqwj3BOixE6AWIm0Kr4y9-ScOm-MKhMs/edit?slide=id.g2f9677cb0f8_0_0#slide=id.g2f9677cb0f8_0_0). However, making this a requirement for an unrelated feature could be seen as using the migration mechanism to enforce an orthogonal best practice.
```

It might be nice to link to some evidence.

dmurph (approved):
LGTM with some optional suggestiongs.

pwa-migration-explainer.md
mkruisselbrink: well, part of this proposal would make these issues not unrecoverable anymore... but yeah, linking to the slides seems fine.

PR #131, opened 2025-12-02
Pwa migration two way handshake

mkruisselbrink:
Taking into account web platform security feedback, update the proposal to require a two-way handshake for cross-origin migrations.


dmurph (approved):
LGTM

PR #136, opened 2026-02-27
Add spec language for PWA Origin Migration

mkruisselbrink:
This adds the initial spec language for PWA Origin Migration to the index.html. Changes include:
- Adding `migrate_from` and `migrate_to` members to the manifest processing.
- Defining the `migrate_from` and `migrate_to` members and their fields.
- Adding the `validate origin migration` algorithm.
- Elevating `web-app-origin-association` to a top-level section.
- Adding non-normative usage examples.
- Adding privacy and security considerations.

***
<a href="https://pr-preview.s3.amazonaws.com/mkruisselbrink/manifest-incubations/pull/136.html" title="Last updated on Mar 11, 2026, 10:28 PM UTC (b926614)">Preview</a> | <a href="https://pr-preview.s3.amazonaws.com/WICG/manifest-incubations/136/2f905dd...mkruisselbrink:b926614.html" title="Last updated on Mar 11, 2026, 10:28 PM UTC (b926614)">Diff</a>

### DIFF
--- a/index.html
+++ b/index.html
@@ -240,6 +240,12 @@
         <li>[=Process the `scope_extensions` member=], passing |json| and
         |manifest|.
         </li>
+        <li>[=Process the `migrate_from` member=], passing |json|, |manifest|
+        and |manifest URL|.
+        </li>
+        <li>[=Process the `migrate_to` member=], passing |json|, |manifest|
+        and |manifest URL|.
+        </li>
       </ol>
     </section>
     <section>
@@ -1780,22 +1786,6 @@
           different use cases.
         </aside>
       </section>
-      <section class="informative">
-        <h3>
-          The web-app-origin-association file
-        </h3>
-        <p>
-          A <dfn>web-app-origin-association</dfn> file is a JSON file that can
-          be used to [=validate scope extensions=]. It confirm an association
-          between the origin it is in with one or more web applications. It
-          identifies apps uniquely by referencing manifest [=manifest/ids=].
-        </p>
-        <p>
-          Given [=origin=] |origin:origin|, a [=web-app-origin-association=]
-          file is expected to be downloadable from
-          <code>[origin]/.well-known/web-app-origin-association</code>.
-        </p>
-      </section>
       <section>
         <h2>
           Validating scope extensions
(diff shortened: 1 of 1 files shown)


index.html
```diff
@@ -1856,6 +1846,350 @@ <h2>
...
+        <ul>
+          <li>
+            <code><dfn data-dfn-for="migrate_to_item">id</dfn></code>: A
+            [=string=] representing the [=manifest/id=] of the new application.
+          </li>
+          <li>
+            <code><dfn data-dfn-for=
+            "migrate_to_item">install_url</dfn></code>: A [=string=]
+            representing a URL of a page that links to the new application's
+            manifest. The manifest on this page needs to include a
```
dmurph, line 1998: I was trying to find something to reference here in the manifest for the 'manifest associated with a document' but couldn't find anything, especially because now it's per-spec to possibly parse a manifest separate from a document. But if you have any ideas of something to reference here that might be nice.

index.html
```diff
@@ -1856,6 +1846,350 @@ <h2>
...
+        <h3>
+          Validating origin migration
+        </h3>
+        <p>
+          To <dfn>validate origin migration</dfn>, given [=URL=]
+          |old_manifest_id:URL| and [=URL=] |new_manifest_id:URL|:
+        </p>
+        <ol class="algorithm">
+          <li>If |old_manifest_id|'s [=URL/origin=] and |new_manifest_id|'s
+          [=URL/origin=] are not [=same site=], return an error.
```
dmurph, line 2014: Should we exit early if they are same-origin?

index.html
```diff
@@ -1856,6 +1846,350 @@ <h2>
...
+          </li>
+          <li>Let |base_origin:origin| be |manifest|["start_url"]'s
+          [=URL/origin=].
+          </li>
+          <li>Let |id_url:URL| be the result of [=URL Parser|parsing=]
+          |json|["id"] with |base_origin| as the base URL.
+          </li>
+          <li>If |id_url| is failure, return.
+          </li>
+          <li>If |id_url| is not [=same origin=] as |manifest|["start_url"],
```
dmurph, line 2064: > thinking

We're kind of re-doing the id parsing, I wonder if we can somehow reference the manifest output from parsing the non-extension manifest members? e.g.:

- take the json manifest AND the parsed manifest.
- If there is no 'id' field in the json, return
- Then, we can just use & trust the id field in the parsed manifest, and not worry about same-origin etc?

Or... if that's what you're doing, why re-parse the id here....

Ok, so this is because the [id parsing algorithm](https://w3c.github.io/manifest/#dfn-process-the-id-member) sets the member and doesn't, say, return a value.

I guess it would be simpler to refactor the manifest spec to have an algorithm return the id or failure, and then have the caller initially set the id to the start_url, and then call that, and only set id if it isn't a failure.

> /thinking

It would be nice not to have to re-implement id parsing, but I see why this is done here. Maybe file an issue afterwards to refactor the manifest spec a little to make this easier.

I _believe_ that the URL Parser >> parse method will always make the result the same origin as base_origin, so the same-origin check isn't necessary.

dmurph (approved):
Generally LGTM. The `id` parsing reimplementation is unfortunate, but we can clean that up in follow-ups.

index.html
mkruisselbrink: not sure either, will update if we come up with something better to reference...

index.html
mkruisselbrink: good catch, yeah, that was missing

index.html
mkruisselbrink: Yeah, re-doing the id parsing is unfortunate. In chrome's implementation we have a "did this manifest have a explicit id" field, maybe the manifest spec could also expose something like that? We do have access to json and parsed manifest here, but I don't think we can treat the presence or absence of id in the json as being good enough to decide if the manifest had an explicitly specified manifest. The parsed manifest will always have a valid id, even if the manifest did specify an id, but that id just happened to be cross origin or something. So if we actually want to guard this on "manifest had an explicit valid id" that should probably be something that the main manifest spec exposes. And until it does, duplicating the id parsing logic here is about the best we can do.

We do need the same origin check as well, since id could be an absolute URL; in which case the origin of the passed in base url doesn't matter.

PR #137, opened 2026-04-09
Add manifest scope improvements doc

LuHuangMSFT:
Describe problems app developers face with the current manifest scope model and propose a two-layer scope architecture to address them.

### DIFF
--- a/manifest-scope-improvements/manifest-scope-improvements.md
+++ b/manifest-scope-improvements/manifest-scope-improvements.md
@@ -28,7 +28,7 @@ This document describes the concrete problems app developers are hitting and pro
 
 **Why it matters.** Navigation capturing is an all-or-nothing proposition today: it applies to every URL inside scope. Developers cannot differentiate between "this URL is part of my app" and "this URL should pull the user into the app window when clicked externally." These are distinct intents, and conflating them forces developers into workarounds (e.g., moving content to a different origin).
 
-**What developers need.** A mechanism to define a **capture scope** that is a subset of (or different from) the application scope — controlling which URLs trigger navigation capturing independently of which URLs render in the app window.
+**What developers need.** A mechanism to define a **capture scope** that is different from the application scope — controlling which URLs trigger navigation capturing independently of which URLs render in the app window.
 
 ---
 
--- a/manifest-scope-improvements/manifest-scope-improvements.md
+++ b/manifest-scope-improvements/manifest-scope-improvements.md
@@ -69,7 +69,7 @@ Design constraints:
 
 ### Layer 2 — In-App Behavior (fine-grained)
 
-This layer answers: *"Now that we know this URL is part of the app, how should the browser handle it?"* This is where richer pattern matching (potentially `URLPattern`) becomes appropriate, because evaluation happens inside the browser after the OS has already routed the URL to the app.
+This layer answers: *"Now that we are in the app, how should the browser handle this URL?"* This is where richer pattern matching (potentially `URLPattern`) becomes appropriate, because evaluation happens inside the browser after the OS has already routed the URL to the app.
 
 **Prior art.** The [`tab_strip.home_tab.scope_patterns`](https://wicg.github.io/manifest-incubations/#home_tab-member) feature in the Manifest Incubations spec already uses `URLPattern` to define which URLs belong to a tabbed app's home tab vs. regular tabs. This demonstrates that URL-pattern-based behavior differentiation *within* a manifest is already a proven pattern, and the work proposed here can build on that precedent.
 
--- a/manifest-scope-improvements/manifest-scope-improvements.md
+++ b/manifest-scope-improvements/manifest-scope-improvements.md
@@ -1,0 +1,107 @@
+# Web App Manifest Scope: Problems and Proposed Improvements
+
+**Author:** Lu Huang (luhua@microsoft.com)
+
+## Introduction
+
+The web app manifest `scope` member defines which URLs belong to an installed web application. Today, scope is a single URL prefix — e.g., `"scope": "/app/"` — and every browser behavior tied to "is this URL part of the app?" inherits that single boundary. This was adequate when scope only governed display-mode switching (standalone vs. browser chrome), but as the platform adds more scope-dependent features — **navigation capturing**, **launch routing** — a single prefix is no longer expressive enough for real-world applications.
+
+The [`scope_extensions`](https://wicg.github.io/manifest-incubations/#scope_extensions-member) feature — now shipping in Chromium — addressed the multi-origin dimension of this problem by letting an app extend its scope to additional origins. When `scope_extensions` was designed, fine-grained same-origin scope control (path exclusions, URL pattern filtering) was [explicitly listed as a non-goal](https://github.com/WICG/manifest-incubations/blob/gh-pages/scope_extensions-explainer.md#non-goals) to keep the proposal focused. With `scope_extensions` shipped, the time is right to tackle these harder problems. The `scope_extensions` explainer's ["Future work"](https://github.com/WICG/manifest-incubations/blob/gh-pages/scope_extensions-explainer.md#future-work-under-consideration) section anticipated this, suggesting include/exclude lists and URL patterns as next steps.
+
+This document describes the concrete problems app developers are hitting and proposes areas of specification work to address them. Specific proposals for each area will be added in their own documents alongside this one.
+
+---
+
+## Problem 1: Scope Cannot Exclude Paths
+
+**The pain.** Many web apps set `"scope": "/"` because their application routes span the root of their origin. However, the same origin may also host content that is *not* part of the app experience — marketing pages, documentation, blog posts, or API endpoints. Today there is no way to say "my app is everything on this origin *except* `/blog/` and `/docs/`."
+
+**Why it matters now.** Before navigation capturing, the consequence was minor: out-of-place pages simply appeared in the standalone window with app chrome. With navigation capturing, clicking a link to `/blog/post-1` on any surface (email, chat, another browser tab) *launches the installed app*, pulling the user into an app window for content the developer never intended to be an app experience. Developers are forced to either shrink their scope (breaking deep links to real app pages) or accept that non-app content appears in the app window.
+
+**What developers need.** A way to *exclude* specific paths or path patterns from scope, so that scope reflects the true boundary of the application.
+
+---
+
+## Problem 2: Navigation Capturing Scope Cannot Differ from App Scope
+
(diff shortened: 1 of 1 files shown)


manifest-scope-improvements/manifest-scope-improvements.md
```diff
@@ -0,0 +1,107 @@
...
+4. **Progressive enhancement.** Apps that don't need fine-grained control should not need to add any new manifest members. The defaults should match today's behavior.
+
+---
+
+## Summary
+
+| # | Problem | Proposed Solution Area |
+|---|---------|----------------------|
+| 1 | Cannot exclude paths from scope | New manifest members for scope include/exclude lists |
+| 2 | Navigation capture scope ≡ app scope | Separate "capture scope" concept (Layer 2) |
```
dmurph, line 99: I wonder if 2,3,4 could be combined into one thing?

E.g.:

```json
"launch_handlers": [
  // Patterns are evaluated in order, until one applies to the url.
  { "url_pattern": <..pattern..>,
     "client_mode": "focus-existing" },
  { "url_pattern": <...pattern..>,
    "client_code": "never_capture" },
  ...
// the pattern could be for exteneded urls! so we could say that an
// extended scope, for example, never captures.

  // to maintain backwards compatibility, if none apply, the "launch_handler" entry is used as the default
],
```

IDK if this is better - to NOT capture something into the app, you have to specify a handler for it.

Questions here would be for backwards compatibility in general too - all solutions need to make sure that old versions or chrome don't break - e.g. the manifest parser doesn't 'throw away' the manifest if it fails to parse something here (I don't think that's the case for most of this)

manifest-scope-improvements/manifest-scope-improvements.md
```diff
@@ -0,0 +1,107 @@
...
+| 3 | Single launch behavior for all URLs | URL-pattern-keyed launch handlers (Layer 2) |
+| 4 | `scope_extensions` lacks feature annotations | Per-origin opt-in/out in association file |
```
dmurph, line 101: Maybe another way to say what I said above is - maybe we can flatten the concept of the per-origin opt-in and opt-out into the same concept of specializing launch handling per url, by introducing a 'never-capture' value? But maybe there are reasons to not do that?

dmurph (approved):

manifest-scope-improvements/manifest-scope-improvements.md
LuHuangMSFT: Interesting. I've not thought about specifying capturing scope from `launch_handlers` before.  Will investigate.

LuHuangMSFT: Do you think this current doc is solution-agnostic enough to be merged?

manifest-scope-improvements/manifest-scope-improvements.md
```diff
@@ -0,0 +1,107 @@
...
+- Must perform well on low-end devices (no regex evaluation at navigation time).
+- Keep the existing `scope` member as-is for backward compatibility; introduce new top-level manifest members (e.g., `scope_exclusions` or an enhanced `scope` object) to add or remove paths.
+
+### Layer 2 — In-App Behavior (fine-grained)
+
+This layer answers: *"Now that we know this URL is part of the app, how should the browser handle it?"* This is where richer pattern matching (potentially `URLPattern`) becomes appropriate, because evaluation happens inside the browser after the OS has already routed the URL to the app.
```
mhochk: If I'm following correctly, we're also talking about this 'layer' being where we think about a clicked link within the app, which we don't know yet if "this URL is part of the app". Small wording tweak suggestion:

```suggestion
This layer answers: *"Now that we are in the app, how should the browser handle this URL?"* This is where richer pattern matching (potentially `URLPattern`) becomes appropriate, because evaluation happens inside the browser after the OS has already routed the URL to the app.
```

manifest-scope-improvements/manifest-scope-improvements.md
```diff
@@ -0,0 +1,107 @@
...
+
+---
+
+## Problem 3: Launch Behavior Cannot Vary by URL
+
+**The pain.** The `launch_handler` manifest member lets a developer choose a `client_mode` — `navigate-new`, `navigate-existing`, `focus-existing` — but it applies uniformly to all launches. A mapping app might want `/map/` navigations to reuse an existing window (`navigate-existing`) while `/trip/new` should open a fresh window (`navigate-new`). There is no way to express this today.
+
+**Why it matters.** Different areas of an app have different interaction models. Forcing a single launch behavior leads to UX compromises: either users lose context when an existing window is reused for unrelated content, or they accumulate unnecessary windows when every launch opens a new one.
+
+**What developers need.** The ability to specify **multiple launch handlers**, each associated with a set of URL patterns, so launch behavior can be tailored to the content being opened.
```
mhochk, line 41: Is URL scoping *enough*? I'm wondering if the app might want to factor in even more dynamic information, like "what URL is my current app page on"? (e.g. YouTube PWA wants to open a video link in the current window if nothing is playing, but if something is playing wants to pause it and open a new window).

If so, this might be more wholistically solved by giving the app more permissions within the context of its launch handler. Namely, letting it easily launch a new window of itself with params similar to the LaunchParams it received.

manifest-scope-improvements/manifest-scope-improvements.md
```diff
@@ -0,0 +1,107 @@
...
+
+### Layer 1 — OS-to-App Scope (coarse-grained)
+
+This layer answers: *"Does this URL belong to the application?"* It determines install-time registration with the OS, deep-link routing, and display-mode application. Because it must map to OS-level URL filtering (Android intent filters, Windows protocol/URI handling, etc.), it should remain **simple** — path prefixes, origin lists, and at most basic include/exclude patterns.
+
+Design constraints:
+- Must be expressible using primitives that operating systems support for URL filtering. Prior work on the (now-obsolete) [`pwa-url-handler` explainer](https://github.com/WICG/pwa-url-handler/blob/main/explainer.md#os-specific-implementation-notes) documented these OS mechanisms: Android [intent filters](https://developers.google.com/web/fundamentals/integration/webapks?hl=ro#android_intent_filters) via WebAPK, Windows ["Apps for Websites"](https://docs.microsoft.com/en-us/windows/uwp/launch-resume/web-to-app-linking), and iOS/macOS [Universal Links](https://developer.apple.com/ios/universal-links/). All are origin- or prefix-based — none support regex or complex pattern matching — which bounds the expressiveness of this layer.
```
mhochk, line 66: FWIW - A browser could choose to support a more complex syntax at this layer, and attempt to best-effort translate it to the OS it is on, erring on the side of a broader capture, with a fall-back of trying to have it handled in the browser. I wouldn't recommend it for a v1, but worth considering as a potential future while designing the API shape.

So for example, on Windows or iOS/macOS the browser might try to break it into include/exclude paths with basic wildcards, whereas on Android it wouldn't have an 'exclude' option. In any case the browser would always check the URL at launch time, and if it didn't match the more complex syntax provided by the PWA it would launch it in the browser mode instead. (Or if it wanted to be super fancy, it could attempt to look up the default browser and forward it along to that. Should be possible on at least Windows, though likely not on Android/iOS).

manifest-scope-improvements/manifest-scope-improvements.md
```diff
@@ -0,0 +1,107 @@
...
+
+---
+
+## Problem 2: Navigation Capturing Scope Cannot Differ from App Scope
+
+**The pain.** Even when a developer is comfortable with a broad app scope (all pages *should* render in standalone mode), they may not want every in-scope URL to *capture navigations* from outside the app. For example, a productivity suite at `app.example.com` may want its `/dashboard/` and `/editor/` paths to capture navigations but not `/settings/privacy-policy`, which is better opened in a regular browser tab.
+
+**Why it matters.** Navigation capturing is an all-or-nothing proposition today: it applies to every URL inside scope. Developers cannot differentiate between "this URL is part of my app" and "this URL should pull the user into the app window when clicked externally." These are distinct intents, and conflating them forces developers into workarounds (e.g., moving content to a different origin).
+
+**What developers need.** A mechanism to define a **capture scope** that is a subset of (or different from) the application scope — controlling which URLs trigger navigation capturing independently of which URLs render in the app window.
```
mhochk: I would lean into the 'different' rather than 'subset' direction. Though a very contrived case, I could imagine wanting to capture a URL activation, but still show it as 'outside the scope of the app' in the UI. Possibly for something like an old URL location that now attempts to guide you to the new URL, but less aggressively than an automatic redirect.

```suggestion
**What developers need.** A mechanism to define a **capture scope** that is different from the application scope — controlling which URLs trigger navigation capturing independently of which URLs render in the app window.
```

mhochk (approved):

dmurph:
Chat at blinkon:
- Add background sectino on scope extensions, and highlight of the intention of ux impact, app exists cross-origins, but only for user ux / user agent usability
- Maybe roll in scope extensinos mentions on problems 1,2,3, sahing that the solution must also be compatible with extended scope sitiuations
- add 'Other Considersations' section or 'Problems' or whatever section
  - "Scope Extensinos T&S Considations - Can the manifest own these extra decisinos or do the exteneded scopes need to control all scope-related features impacting viewing their urls on the PWA"

proposal from me would likely put all controls in manifest and only delegate to the scope extension the 'scope filtering' part? eg.g the scope member field and any new fields we add to modify scope for exlucsion etc, but then the 'UX behavior' bits with capturing and launch customization etc be in the manifest, and just have it able to specify url patterns or whatever that include extendeds origin. BUT - alternative should be docujented to have this be in the scope extensions dict area. We then should consult with T&S about what they think. Is this blocking? or likely ok?


adding fields to better specify scope in both  manifest and scope extensions dict sounds appropriate. other origins can specify the scope more fine-grained. but the actuall UX controls over launch handlign and capturing IN that scope can not make any situation "worse" for that extended scope origin than today (from a t&s point of view) -it's just making the app more UX-robust. So it seems fine to have all of that in the manifest not need to ,say, put a 'launch_handler' member field in the scope_extensions dict.

manifest-scope-improvements/manifest-scope-improvements.md
LuHuangMSFT: That's a fair comment - the oldest proposal that would be able to dynamically handle launches taking into account the state of all windows is sw-launch (https://github.com/WICG/web-app-launch/blob/main/sw_launch_event.md). 

Thinking about whether we can work this into what's currently available in launch handling - if "focus-existing" client mode is used, the app has the opportunity to then "think" about what to do next using its current state and params from `window.launchQueue`. The state visible here is probably less that what would be available to the service worker.

manifest-scope-improvements/manifest-scope-improvements.md
LuHuangMSFT: Will keep this in mind. We have experience (on Windows at least) implementing such a return-the-URL for a system default browser launch behavior.

LuHuangMSFT:
Will merge first, then incorporate dmurph's feedback for next edit.

PR #139, opened 2026-05-04
Spec: Define the 'unframed' display mode for Isolated Web Apps

edman:
This PR updates the manifest-incubations specification to introduce details of the `unframed` display mode.

Side changes:
- Updated `app-region` CSS property definition to restrict its effects to the `unframed` display mode.
- Linked to the Isolated Web Apps explainer.
- Fixed ReSpec errors and warnings.
- Converted note containers from `<p>` and `<div>` to `<aside>`.

***
<a href="https://pr-preview.s3.amazonaws.com/edman/manifest-incubations/pull/139.html" title="Last updated on May 29, 2026, 10:45 AM UTC (15f1e2e)">Preview</a> | <a href="https://pr-preview.s3.amazonaws.com/WICG/manifest-incubations/139/58db16f...edman:15f1e2e.html" title="Last updated on May 29, 2026, 10:45 AM UTC (15f1e2e)">Diff</a>


edman:
Hi @dmurph, this is my attempt to make the changes we discussed in our last meeting. PTAL

edman:
A couple things I wanted to ask/clarify with you:

* I'm reusing the URL pattern parsing algorithm as done in `scope_patterns` https://wicg.github.io/manifest-incubations/index.html#processing-the-scope_patterns-member. We had discussed to define the algorithm in this spec, but I think it's not needed. Please review that part.

* I see that there is no "Processing the `display_override` member" section or similar to describe the algorithm that parses display override from the manifest. Should we add one? We will need it if we want to ignore non-unframed display overrides with URL patterns in the manifest parsing stage.

* Do we want to ignore any non-unframed usage of the new object type for display override? Or is it acceptable to use the object type for display modes with no URL patterns (null or empty list)? The current approach is to ignore modes if it's not unframed.

index.html
```diff
@@ -206,6 +349,14 @@ <h3>
           <li>To disable dragging an element, shall be set to `no-drag`.
           </li>
         </ul>
+        <p>
+          [=User agents=] MUST restrict the effects of `app-region` to when the applied [=display
+          mode=] is [=display mode/unframed=].
```
dmurph: What does this mean? It sounds like it's only used in unframed, but we use it in window-controls-overlay?

Or - are you saying user agents have to do something? what does 'restrict the effects' mean?

index.html
```diff
@@ -59,55 +67,109 @@ <h2>
...
-        value is a [=display mode=].
+        "appmanifest#dfn-display">display mode</a> for the web application.
       </p>
       <p>
         The [=manifest/display_override=] member of the [=application
-        manifest=] is a <a>sequence</a> of <a data-cite=
-        "appmanifest#dfn-display-modes-list">display mode list values</a>
-        including extensions like [=display mode/window-controls-overlay=] and
-        [=display mode/unframed=]. This member represents the developer's
-        preferred fallback chain for [=display mode=]s. This field overrides
```
dmurph, line 71: Why did you remove the text about overriding here? That seems important?

index.html
```diff
@@ -59,55 +67,109 @@ <h2>
...
+          <a data-cite="appmanifest#dfn-display-modes-list">display modes list
+          value</a> including extensions.
+        </li>
+        <li>
+          <dfn data-dfn-for="DisplayOverrideEntryObject">url_patterns</dfn> (optional): a
+          [=list=] of [=URL patterns=]. Defaults to an [=list/is empty|empty list=] if omitted.
+        </li>
+      </ul>
+      <aside class="note">
+        Currently, a [=DisplayOverrideEntryObject=] is only supported if its
```
dmurph: If you want to enforce this MUST, then you probably need to encorporate this into the parsing algorithm?

OK to keep as a note too.

index.html
```diff
@@ -59,55 +67,109 @@ <h2>
...
+        [=DisplayOverrideEntryObject/display=] member is [=display mode/unframed=]. [=User agents=]
+        MUST ignore object entries with any other display mode.
+      </aside>
       <p>
         The following steps are added to the [=application manifest/processing
         extension-point=] in <a data-cite=
         "appmanifest#dfn-steps-for-determining-the-web-app-s-chosen-display-mode">
-        determining the web app's chosen display mode</a>:
+        determining the web app's chosen display mode</a>, which now also
+        takes an optional [=URL=] |initial URL|:
```
dmurph: maybe document_url?

index.html
```diff
@@ -59,55 +67,109 @@ <h2>
...
       </p>
       <ol>
-        <li>[=list/For each=] |candidate_display_mode:DisplayModeType| of
-        |manifest:ordered map|.[=manifest/display_override=] member:
+        <li>[=list/For each=] |entry:DisplayOverrideEntryObject| of
```
dmurph: it seems like we don't want to declare the type yet - if this is the generic algorithm for processing this, we should be undefined here as type, and then the first step likely is to split the logic depending on if it is a string or dictionary.

index.html
```diff
@@ -59,55 +67,109 @@ <h2>
...
+        |manifest:ordered map|'s [=manifest/display_override=] member:
           <ol class="algorithm">
-            <li>If <a data-cite="appmanifest/#dfn-display-modes-list">display
-            modes list</a> contains |candidate_display_mode:DisplayModeType|,
-            return that |candidate_display_mode:DisplayModeType|
+            <li>Let |candidate_display_mode| be null.
+            </li>
+            <li>If |entry| is a [=display mode=]:
```
dmurph: Is it normal to compare it to a type like this? In my head this is usually saying "if it is a string" AND if it matches a display mode

index.html
dmurph: (see below how you ask if it's a displya mode? in this algorithm, it cannot be - in fact, this iteration we would fail with an error on the first string in this list)

index.html
dmurph: Here is one way:
https://github.com/w3c/manifest/blob/c0d4307a9098f4b54d03fa4b31b16f6e3f683f81/index.html#L700

index.html
```diff
@@ -59,55 +67,109 @@ <h2>
...
+              <ol>
+                <li>Set |candidate_display_mode| to |entry|.
+                </li>
+              </ol>
+            </li>
+            <li>Else if |entry| is a [=DisplayOverrideEntryObject=]:
```
dmurph: same - we would be checking if its a dictionary, etc, and then create the type explicitly based on the json. Maybe there is other places in respec where we just say "if _undefined_type_value_ is _dictionary_type_" - if that happens somewhere let me know.

index.html
```diff
@@ -59,55 +67,109 @@ <h2>
...
+                    </li>
+                  </ol>
+                </li>
+                <li>Else:
+                  <ol>
+                    <li>If |initial URL| is null, [=iteration/continue=].
+                    </li>
+                    <li>[=list/For each=] |pattern| of |entry|'s [=DisplayOverrideEntryObject/url_patterns=]:
+                      <ol>
+                        <li>If |pattern| [=URL pattern/matches=] |initial URL|, set |candidate_display_mode| to [=display mode/unframed=] and [=iteration/break=].
```
dmurph: oh... so you are also implementing the algorithm for 'specify a url and see what display mode the manifest says should occur for it'. I see.

dmurph: Normally, I would recommend this:

- Define an algorithm to 'process' the json of display_override
- Define an algorithm to 'apply' a post-processed manifest object (which includes the display_override information, if you define it saving stuff there) to a given url to decide the display mode for that url.

That way, you can get all of the json parsing minutiae out of the way first to ensure that the value saved on manifest is not invalid, and then you have a separate algorithm for, given a url and manifest, what display mode should it be.

I recommend the initial parsing algorithm 'returning' the first non-frameless display mode, as the default display mode, and then accepting the 'manifest' object in the algorithm and saving all of the DisplayOverrideEntryObject objects on the manifest object's "display_override". Then, the display mode choosing algorithm can treat the manifest's 'display' as the default, and iterate the display_override first, which now MUST be DisplayOverrideEntryObjects, and potentially return one of those given the url

dmurph (changes requested):

index.html
edman: > Define an algorithm to 'process' the json of display_override

Done in "processing the display_override member".

> Define an algorithm to 'apply' a post-processed manifest object (which includes the display_override information, if you define it saving stuff there) to a given url to decide the display mode for that url.

Done as this algorithm, but moved to a new subsection "determining the chosen display mode".

> Then, the display mode choosing algorithm can treat the manifest's 'display' as the default, and iterate the display_override first, which now MUST be DisplayOverrideEntryObjects, and potentially return one of those given the url

IIUC this doesn't work in a case like

```json
{
  "display_override": [
    "minimal-ui",
    { "display": "unframed", "url_patterns": [ ... ] },
  ]
}
```

Here the window should never be unframed because the first minimal-ui override takes precedence. Generally if we split unframed in a separate list and prioritize it, we'll lose the ordering between unframed and other modes, so the window can become unframed when it shouldn't.

index.html
edman: Also, covered with the new algorithm to parse display_override from json?

index.html
edman: I believe this is also covered in the new algorithm to parse display_override from json.

index.html
edman: Is this resolved given we now have a separate algorithm to parse the display_override from json?

index.html
edman: I see other variables in this spec often use names like "manifest URL" instead of "manifest_url". I named this "document URL", let me knnow if I should go with "document_url" instead.

index.html
edman: Done, added the manifest json parsing algorithm with this behavior encoded.

The note is still there for extra clarity, can remove if you think it's redundant.

index.html
edman: Done, I removed it by mistake while changing the member definition. Brought it back in a paragraph below.

index.html
edman: "restrict the effects to unframed" means it should only work in unframed. I rephrased this to make it more direct.

I wasn't sure what to do about WCO. The spec currently says `app-region` is not implemented, so I focused on changing it for unframed. But you're right chrome's implementation supports unframed and WCO. Should I add it here too?

index.html
```diff
@@ -59,55 +67,175 @@ <h2>
...
+        [=display mode/tabbed=], and [=display mode/unframed=]) or a
+        [=DisplayOverrideEntryObject=].
       </p>
       <p>
-        The following steps are added to the [=application manifest/processing
-        extension-point=] in <a data-cite=
-        "appmanifest#dfn-steps-for-determining-the-web-app-s-chosen-display-mode">
-        determining the web app's chosen display mode</a>:
+        This member represents the developer's preferred fallback chain for
+        [=display modes=]. This This field overrides the [=manifest/display=]
```
dmurph:
```suggestion
        [=display modes=]. This field overrides the [=manifest/display=]
```

index.html
```diff
@@ -59,55 +67,175 @@ <h2>
...
+                      </li>
+                    </ol>
+                  </li>
+                  <li>Else:
+                    <ol>
+                      <li>If |document URL| is null, [=iteration/continue=].
+                      </li>
+                      <li>[=list/For each=] |pattern| of |entry|'s [=DisplayOverrideEntryObject/url_patterns=]:
+                        <ol>
+                          <li>If |pattern| [=URL pattern/matches=] |document URL|, set |candidate display mode| to [=display mode/unframed=] and [=iteration/break=].
```
dmurph: The linked manifest spec area of this extension point has some language like:
`If the user agent supports manifest["display"], then return manifest["display"].`

I think.... we have similar language to this below. but here we exit early. I recommend not exiting early, and instead setting the candidate display mode here? that lets us appropriately either continue iterating (as it'll be null) if the url pattern doesn't match, and potentially continue iterating if the user agent doesn't support unframed.

index.html
```diff
@@ -59,55 +67,175 @@ <h2>
...
-            <li>If |candidate_display_mode:DisplayModeType| is [=display
-            mode/tabbed=] and the user agent supports this, then return that
-            |candidate_display_mode:DisplayModeType|.
-            </li>
-          </ol>
+      <ul>
+        <li>
+          <dfn data-dfn-for="DisplayOverrideEntryObject">display</dfn>: a
+          <a data-cite="appmanifest#dfn-display-modes-list">display modes list
+          value</a> including extensions.
```
dmurph: here (and elsewhere) it might be nice to link to the extensions.

Especially in the algorithm - linking to the display mode extensions header would be helpful.

index.html
```diff
@@ -59,55 +67,175 @@ <h2>
...
+          </li>
+          <li>If |json|["display_override"] is not a [=list=], return.
+          </li>
+          <li>Let |processed display override:list| be a new [=list=].
+          </li>
+          <li>[=list/For each=] |entry| of |json|["display_override"]:
+            <ol>
+              <li>If |entry| is a [=string=]:
+                <ol>
+                  <li>If the <a data-cite="appmanifest#dfn-display-modes-list">display modes list</a> (including extensions) contains |entry|, [=list/append=] |entry| to |processed display override|.
```
dmurph: (link extensions to the list below)

index.html
```diff
@@ -59,55 +67,175 @@ <h2>
...
+                  </li>
+                </ol>
+              </li>
+              <li>Else if |entry| is an [=ordered map=]:
+                <ol>
+                  <li>If |entry|["display"] does not [=map/exist=] or is not a [=string=], [=iteration/continue=].
+                  </li>
+                  <li>Let |display mode| be |entry|["display"].
+                  </li>
+                  <li>If the <a data-cite="appmanifest#dfn-display-modes-list">display modes list</a> (including extensions) does not contain |display mode|, [=iteration/continue=].
```
dmurph: (link extensions to the list below)

index.html
```diff
@@ -59,55 +67,175 @@ <h2>
...
+                  </li>
+                </ol>
+              </li>
+            </ol>
+          </li>
+          <li>Set |manifest|["display_override"] to |processed display override|.
+          </li>
+        </ol>
+        <aside class="note">
+          Currently, a [=DisplayOverrideEntryObject=] is only supported if its
```
dmurph: Ok I thinka good way to simplify this a little is like so:
- The algorithm for parsing can have text for parsing the dictionary like: "If the user agent doesn't support the display mode with url patterns, continue"
  - This means that, display_override on the manifest cannot have a dictionary entry unless the user agent supports url patterns for that display mode. this limits that, then, to just unframed. And we don't care if the url_patterns field is populated or not - no matter what, we only support dictionary-form if the user agent supports url_patterns for that display mode.
- Then, the 'apply a display mode' algorithm doesn't have to care about unframed at all - it can just assume that if a dictionary entry is there, it's for a display mode that the user agent supports url_patterns on. Otherwise it wouldn't have parsed.

Then, we can remove this specific language around unframed and MUST - but we **can** write something like - "After processing, the only dictionary entries that can exist in the manifest's display_override are for display modes that the user agent supports url_patterns for." - a better phrased maybe version of that.

WDYT?

index.html
```diff
@@ -184,17 +319,91 @@ <h4>
...
+            controls. In contrast, the unframed app window (right) removes these host-provided frame
+            decorations, allowing the web contents to occupy the entire window area.
+          </p>
+          <div style="display: flex; gap: 20px; margin-top: 15px;">
+            <figure style="margin: 0;">
+              <img src="images/example-non-unframed-window.png" alt="Example regular app window" style="max-width: 100%; height: auto;">
+              <figcaption style="text-align: center; font-size: 0.9em; color: #555;">Regular app window in standalone mode</figcaption>
+            </figure>
+            <figure style="margin: 0;">
+              <img src="images/example-unframed-window.png" alt="Example unframed app window" style="max-width: 100%; height: auto;">
```
dmurph, line 359: this is broken in the preview but it might work once committed?

index.html
```diff
@@ -184,17 +319,91 @@ <h4>
...
+          web contents to create a customized title bar.
+        </p>
+        <p>
+          [=User agents=] MUST restrict the use of [=display mode/unframed=] to
+          [[[isolated-web-apps]]].
+        </p>
+        <p>
+          Once an [=display mode/unframed=] window has been created, its display
+          mode is fixed for the lifetime of the window. Navigation to out-of-scope
+          URLs MUST result in a new, regular window being opened for the
```
dmurph: what does 'regular' mean? and is this a browser window? browser tab? or iwa window?

dmurph (changes requested):

index.html
edman: Done, thanks

index.html
edman: I removed specific mentions to unframed, and removed the `continue` calls within the "if entry is an object" condition.

PTAL.

index.html
edman: Done.

WDYT about making a new definition for "extended display modes list values"? We could use that directly instead of the current phrasing with "display modes list value (including extensions)".

index.html
edman: Done

index.html
edman: Done

index.html
edman: Sounds good on my side, I've made those changes.

Is it ok not to mention that unframed supports URL patterns? Maybe it is, since this wording leaves it up to the user agent to decide and implement URL patterns for any display mode it wants.

index.html
edman: It seems pr-preview doesn't load assets https://github.com/tobie/pr-preview/issues/87

It works on my fork https://edman.github.io/manifest-incubations/#visual-example

index.html
edman: The requirement is that out of scope navigation cannot happen in an unframed window. I changed the wording to convey that more directly:

> The user agent MUST NOT allow navigation to out-of-scope URLs to take place within the unframed window.

In practice in Chrome this will be a new tab, or a new PWA window depending on link capturing. You can't link to an IWA, so this can't be an IWA window.

index.html
```diff
@@ -59,55 +67,169 @@ <h2>
...
+      <section>
+        <h3>
+          Processing the `display_override` member
+        </h3>
+        <p>
+          To <dfn>process the `display_override` member</dfn>, given [=ordered map=]
+          |json:ordered map|, [=ordered map=] |manifest:ordered map|, and [=URL=]
+          |manifest URL:URL|, run the following:
+        </p>
+        <ol class="algorithm">
```
dmurph, line 178: nit: Add an initial step of setting the manifest's display override member to an empty list, so an early return still has it populated (so the 'determining' algorithm still 'works').

dmurph (approved):
approved with one nit!

index.html
edman: Done, thanks for catching that

Issue #105, opened 2024-07-13
Better fine-grained 'scope' control (in normal manifest scope & scope extensions)


jacobamills:
In [scope_extensions-explainer.md#future-extensions](https://github.com/WICG/manifest-incubations/blob/gh-pages/scope_extensions-explainer.md#future-extensions) the following is mentioned as a possibility:

> - More specific scoping e.g. scope suffix or include/exclude lists or [URL patterns](https://wicg.github.io/urlpattern/).
>     - To be able to apply these more specific scoping rules to the primary scope (including exclusion). One possible approach is to have the primary origin specified in the scope_extensions list and have it override the behaviour of scope.

For Google Meet, this is a blocker for us to be able to enable link capturing, as there are specific paths within our application that we do not want our PWA to handle.

Because our meeting codes are defined on the root path (e.g. meet.google.com/aaa-bbbb-ccc), we would specifically need path exclusion, rather than just inclusion. For example, we would want to be able to:
- include: meet.google.com/aaa-bbbb-ccc
- exclude: meet.google.com/about

jacobamills:
@dmurph Thank you for reaching out to me about this proposal

@LuHuangMSFT FYI

dmurph:
Amazing, thank you for this use-case! This is incredibly valuable.

@LuHuangMSFT I think this matches with our previous convo. Notes are [here](https://docs.google.com/document/d/16g8_477v2vZ6fbfYQVabuY0Lavy93MvSnVisMDCcgg8/edit#heading=h.wwo73bcn8z7g)

General conclusion currently:
- The above use-case and general behavior seems to be wanted for all scopes, not just extensions or the 'manifest scope'
- So to solve this, we will have a separate 'exclusions' field that is used on top of the manifest scope AND the scope extensions.
- We looked into per-platform capabilities (like Android and iOS) and something like url pattern seems totally incompatible with them. However, prefix-based solutions seem supported. prefix-based exclusion is supported on iOS, but not (yet) on Android.

Something kinda like this is the conclusion:

```json
"scope": "https://www.example.com/app/",
"scope_extensions": [
       { "type": "origin", "value": "https://helpcenter.example-help-center.com" }
]
"scope_exclusions": [
  // This apples to all origins, so this removes any url with the prefixes:
  // https://www.example.com/app/scope/
  // https://helpcenter.example-help-center.com/app/scope/
  "/app/about/",

  // These only apply to the specific origins
[...]

PR #143, opened 2026-07-07
Add scope_inclusions / scope_exclusions explainer

LuHuangMSFT:
Adds a detailed explainer for two proposed Web App Manifest members —
**`scope_inclusions`** and **`scope_exclusions`** — as a companion document
under `manifest-scope-improvements/`.

### What
- **`scope_exclusions`** removes specific paths/patterns from an app's
  navigation scope (e.g. keep `/logout`, `/about`, `/api/*` out of the app window).
- **`scope_inclusions`** adds non-contiguous or parameterized paths that a single
  `scope` prefix can't express (e.g. `/users/:id/dashboard`).
- Both are **order-independent lists** (exclude-wins), drawn from a restricted,
  OS-mappable subset of [URLPattern](https://urlpattern.spec.whatwg.org/).

### Why
This fills in the "own document" that the
[manifest-scope-improvements overview](https://github.com/WICG/manifest-incubations/blob/gh-pages/manifest-scope-improvements/manifest-scope-improvements.md)
anticipates for **Problem 1 (Scope Cannot Exclude Paths)**, plus its include-side
counterpart. Both members live entirely in the overview's **Layer 1** (coarse,
OS-mappable scope); Layer 2 in-app behaviors are out of scope here.

### Notes
- **Explainer only** — no normative/spec changes.
- **Backward compatible**: `scope` is unchanged; the new members are additive.
- Alternatives (single ordered list, prefix-only, replace-vs-union) are covered
  in the doc with rationale for the two-list choice.

Grounded in developer reports WICG/manifest-incubations#105 and w3c/manifest#996.
Related: Chromium issue https://issues.chromium.org/issues/380489997


manifest-scope-improvements/scope-inclusions-and-exclusions.md
```diff
@@ -0,0 +1,315 @@
...
+- Regular expression groups (may be restricted for performance/security, per §4 of URL Pattern spec)
+- Replacing or deprecating the `scope` member
+- **Ordered, position-dependent matching and re-inclusion** (excluding a subtree then re-including a leaf) — deliberately out of scope; see [Alternatives considered](#alternatives-considered)
+- Platform-specific scoping APIs (Android App Links, iOS Universal Links) — those remain independent mechanisms
+
+## Proposed Approach
+
+Introduce two new manifest members that refine the set of in-scope URLs, leaving the existing `scope` member unchanged:
+
+- **`scope_exclusions`** — a list of URL patterns to *remove* from scope; the primary, urgently-requested capability. Exclusions apply **globally**: they may remove paths from the app's own origin and from any origins the app adds via `scope_extensions`. This needs no other origin's consent, because removing a URL from scope only ever shrinks the app's reach.
```
dmurph: Question - can we have it only apply to the current origin of the json? I think this was part of the reason we scoped the scope_extensions json to match the one here - that way we can have these appaer in there too.

e.g. - in scope_extensions, the .well-known file can specify a "scope" member, matching the manifest algorithm for in-scope. So by adding these fields, we can have them apply in the same in-scope algorithm for a scope field and a url, and then when we extend that algorithm for scope extensions, we can re-use the same algorihtm.

I'm not describing this very well... but do you know what I mean?

LuHuangMSFT: I'm flexible on whether we (A) put all exclusions in the manifest, or (B) have the parallel structure (scope, scope_inclusions, scope_exclusions) in the manifest and WAOA files. I remember at one point we thought it was a good idea to place all exclusions in the manifest. 

Agree that it'd be easier to apply the same algorithms if (B).

dmurph: @mkruisselbrink WDYT? @christianliebel WDYT?

And @jacobamills WDYT? 

(from convo here: https://github.com/WICG/manifest-incubations/issues/105)

dmurph: My intuition here for using scope extensions file:
- slightly harder to update in some ways (multiple files), but easier in other ways (don't have to update manifest file to change scope on an extended origin)
- more secure - the configuration for origin X is hosted on a file on origin X

I'm leaning towards putting the scope extension configuration in the .well-known origin association file. But let me know if I missed something

LuHuangMSFT: Edited to have `scope`/`scope_inclusions`/`scope_extensions` work on their own origin (from main manifest, or from WAOA.) Good to merge?

